PR #9, opened 2015-12-15
refs #50230 - if there are no roles, then create, rather than deletin…

LeChin:
…g then creating. deleting just starts the count higher and doesnt truncate the table

;


jamesmk:
:+1: Can you add a changelog item?

thanks!

PR #16, opened 2016-01-11
refs #48759 - Stack filter buttons

tshedor:
Clean up the mess that was the filter group CSS and allow filter buttons to stack instead of collide with filter selects


jamesmk:
@tshedor: Looks good, can you add a changelog item?

@markhoffmann or @kyleatfine: can we approval on the CSS?

kyleshevlin:
Looks fine to me.

markhoffmann:
i love it

PR #17, opened 2016-01-16
added in helper text as an option for nested_table

nilesvm:


nilesvm:
Roger you got it haus :)

PR #18, opened 2016-01-19
Content form options

nilesvm:


app/views/fae/application/_content_form.html.slim
```diff
@@ -6,6 +6,9 @@ ruby:
   markdown           ||= false
   markdown_supported ||= false
   hint               ||= nil
+  wrapper_html       ||= {}
+  input_options      ||= {}
+  input_html         ||= {}
```
jamesmk: Can you add this local as a method argument for `fae_content_form` and document or remove it here?

app/views/fae/application/_content_form.html.slim
```diff
@@ -15,8 +18,12 @@ ruby:
...
-= f.simple_fields_for attribute.to_sym do |i|
-  - input_options = { data: { validate: i.object.validation_json(f.object.class) }}
-  - input_options[:class] = 'js-markdown-editor' if markdown.present?
-  = i.input :content, label: label_adjusted.html_safe, hint: hint, input_html: input_options
+  options[:input_html].merge! class: 'js-markdown-editor' if markdown.present?
+
+  options.merge! input_options if input_options.present?
+
+= f.simple_fields_for attribute.to_sym, defaults: { wrapper_html: wrapper_html } do |i|
+  - options[:input_html].merge! data: { validate: i.object.validation_json(f.object.class) }
```
jamesmk: I like how you pulled out the options into their own hash. Can you move this line up so all the option munging is happening in one spot?

jamesmk:
A dummy use case and tests on the new options would be nice, if you had time.

app/views/fae/application/_content_form.html.slim
nilesvm: @jamesmk that last option addition needs to be with in the context of simple_fields_for I believe.

PR #23, opened 2016-02-07
refs #54606 update inject_into_file :after for initial routes.rb setu…

kyleatfine:
…p to catch more possible variations of file's opening line


jamesmk:
:+1:

PR #29, opened 2016-02-22
refs #53749 - don't show filters when records are not present

tshedor:


jamesmk:
@tshedor I'm seeing some failing tests on Shippable, are they passing locally for you?

PR #30, opened 2016-02-24
54625 asset initializer

jamesmk:
Note: also tangled up in here is deploy cleanup after making `slim` a dependency


tshedor:
I approve

PR #35, opened 2016-02-26
54683 titlize

LeChin:


spec/dummy/db/schema.rb
```diff
@@ -144,6 +144,18 @@
 
   add_index "fae_options", ["singleton_guard"], name: "index_fae_options_on_singleton_guard", unique: true, using: :btree
 
+  create_table "fae_pages", force: :cascade do |t|
```
jamesmk: I think you have a relic in your local DB. Can you verify you have a `fae_static_pages` table and delete the `fae_pages` table. When you migrate again, this should go away.

PR #45, opened 2016-03-10
52556 clone confirm

tshedor:


jamesmk:
@tshedor fix the conflicts and we're good to go.

PR #47, opened 2016-03-12
50420 customcolumns

LeChin:


jamesmk:
@LeChin can you add a couple tests verifying custom `attr` and `title` coming through?

thanks!

PR #54, opened 2016-03-18
55907 nav breadcrumb updates

jamesmk:


jamesmk:
Looks like I'm pulling in commits that are already on v1.3, let me reopen...

PR #56, opened 2016-03-21
refs #55864 - add fae_avatar for gravatar support

tshedor:
This implementation adds a view helper. I originally added an `avatar:string` to `Fae::Users` but I realized that might be a bit more heavy-duty than what this feature calls for (it would update the column after every successful authentication).

The downside of the helper implementation is calculating the MD5 on every request. I don't think this is significant enough to be a performance concern, but, you know, every millisecond can add up. @jamesmk let me know if you'd prefer the database method where MD5 only runs once per login.


jamesmk:
Looks good, can you add a changelog item?

PR #73, opened 2016-04-05
55178 nav rework

jamesmk:
@tshedor this fixes a couple nav bugs


app/models/fae/navigation.rb
```diff
@@ -58,7 +58,7 @@ def increment_coordinates
     end
 
     def current_section
-      @current_path.gsub(/\/new|\/[0-9]\/edit/, '')
+      @current_path.gsub(/\/new|\/[0-9]*\/edit/, '')
```
tshedor: Do you mean `*` or `+` ? `*` is going to include characters not numeric. Alternately, `\d+`

jamesmk: Wasn't to concerned as integers would only be used as IDs in the URL, but specificity never hurts. Updated.

PR #77, opened 2016-04-07
remove outdated crop feature

tshedor:
Since we're not actively supporting this feature, it should be removed. Otherwise, it's just cruft and misleading.

This commit can be reversed if we eventually want to support crop, but reverter be aware - jcrop.js will need to be replaced or readded


jamesmk:
Love it, look at all those deletions! :heart:

PR #84, opened 2016-04-11
refs #56439 - consolidate migrations

tshedor:
Although I know it's not best practice to do this, errors in change_field_in_fae_text_fields (rename_column also renames the index in Rails >= 4.2) motivated this commit. Since Fae does not yet have widespread adoption, I think this change can be absorbed without consequences. The Rails engine only copies in new files, and this commit does not rename any existing files.

Fae versions < 1.1 should be upgraded to v1.1 and run `rake fae:install:migrations` before upgrading to > 1.1 as this commit consolidates the language feature migration (8150a84).

For reference, [this gist has the schema](https://gist.github.com/tshedor/143156714c111fc3839de16cfe2046d2) generated after `rails g fae:install` for the [consolidated schema](https://gist.github.com/tshedor/143156714c111fc3839de16cfe2046d2#file-consolidated-schema-rb) (the changes made in this PR) vs. the [original, unconsolidated schema](https://gist.github.com/tshedor/143156714c111fc3839de16cfe2046d2#file-pre-consolidated-schema-rb). The only difference is the order of the columns as they are listed within tables, which is irrelevant. Schema generated with Rails 4.2.6 on MySQL.


docs/upgrading.md
```diff
@@ -16,6 +16,14 @@ $ rake db:migrate
 
 ---
 
+# From v1.0 to v1.3
```
jamesmk: I don't think this is needed as consolidating these migrations shouldn't affect anyone and this could potentially cause confusion.

docs/upgrading.md
```diff
@@ -16,6 +16,14 @@ $ rake db:migrate
...
+
+Fae v1.3 consolidates install migrations. When upgrading from v1.0 to v1.3 <=, Fae will have to be [upgraded to v1.1 and have the new migrations copied](#from-v10-to-v11):
+
+After a successful migration, follow any [additional upgrading steps](#from-v11-to-v12) before proceeding. For more information, please see [the migration consolidation PR](https://github.com/wearefine/fae/pull/84).
+
+---
+
 # From v1.0 to v1.1
```
jamesmk: You can delete this section as you removed the migration needed to make this upgrade.

jamesmk:
If Fae had been in the wild longer, this wouldn't have worked as it breaks potential upgrading. However, we don't have any installs < v1.1, so I think we're safe here.

tshedor:
Yeah, I agree. Updated.

docs/upgrading.md
jamesmk: Delete the "From v1.0 to v1.1" section and we're good here.

thanks

tshedor: FWIW, I didn't add "From v1.0 to v1.1" in this PR, but I deleted it.

PR #89, opened 2016-04-12
Hotfix activity - Updated change_item_link method so that it works w fae_display_fields that are integers not just strings, and the /admin/activity page works

nilesvm:
I'll swing back to this one after launch, probably with the updates to nested_table paths.

### DIFF
--- a/app/helpers/fae/application_helper.rb
+++ b/app/helpers/fae/application_helper.rb
@@ -98,7 +98,9 @@ module Fae
       if change.changeable_type.exclude?('Fae') && change.changeable_type.exclude?('Page') && (!ActiveRecord::Base.connection.table_exists? change.changeable_type.tableize)
         return "#{change.changeable_type}: model destroyed"
       else
-        text += change.try(:changeable).try(:fae_display_field).to_s || "##{change.changeable_id}"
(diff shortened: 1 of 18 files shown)


tshedor:
@nilesvm Changelog item?

app/helpers/fae/application_helper.rb
```diff
@@ -98,7 +98,7 @@ def change_item_link(change)
       if change.changeable_type.exclude?('Fae') && change.changeable_type.exclude?('Page') && (!ActiveRecord::Base.connection.table_exists? change.changeable_type.tableize)
         return "#{change.changeable_type}: model destroyed"
       else
-        text += change.try(:changeable).try(:fae_display_field) || "##{change.changeable_id}"
+        text += change.try(:changeable).try(:fae_display_field).to_s || "##{change.changeable_id}"
```
jamesmk: @nilesvm this will break existing functionality. Previously if `change.try(:changeable).try(:fae_display_field)` was `nil` it could use the `change.changeable_id`. But `nil.to_s == ''` so it would never fallback to the ID.

```
>   nil || "something"
 => "something"
> nil.to_s || "something"
 => ""
```

Unfortunately, you'll need to update this to use a conditional.

PR #90, opened 2016-04-12
refs #56749 - highlight the nav with navigation coordinates

tshedor:
Highlight nav elements using the new `@fae_navigation.coordinates` and add/reactivate some tests


app/views/fae/application/_header.slim
```diff
@@ -3,10 +3,10 @@ header.main-header#js-main-header
   h1: a href=root_path = @option.title
 
   - if @fae_topnav_items.present?
-    nav.js-nav
-      ul.main-header-nav
+    nav
+      ul.main-header-nav.js-nav data-coordinates=@fae_navigation.coordinates[0..1].join(',')
```
jamesmk: Instead of getting JS involved, why not just use the index of the loop and compare it to the coordinate of the corresponding level? If they are the same you add the `-current` class via Ruby.

spec/features/nav_spec.rb
```diff
@@ -2,66 +2,76 @@
 
 feature 'Main Navigation' do
 
-  # TODO: fix failing specs
+  scenario 'should highlight first and second level in header', js: true do
```
jamesmk: I added some similar tests in the nav PR I have open (that I'm failing to update). Let's get that merged and see where these overlap.

app/views/fae/application/_header.slim
tshedor: It made the Slim look super gross, and the JS solution was less lines. In the Slim solution, I found I was writing the same code over and over again, but it was too small (ternary here, interpolation there) to modularize. 

If you prefer this to be a Slim solution, I can go back to that.

jamesmk: Hmm, you weren't able to package it into a helper? For me, the logic is harder to follow when it's split between Ruby and JS and it should be slightly more performant if it's Ruby-only. I would prefer that route.

tshedor:
@jamesmk I tried to put this into a helper method, I really did. But the code quality is lacking and it doesn't work. I'm in favor of the JS method because it traverses the DOM much better and it factors in active classes being on both `li` and `a` tags.

The classes could be applied to `a` tags exclusively, but then the CSS code quality/readability will suffer.

jamesmk:
@tshedor so you're proposing to revert what you've done here in favor of the JS method?

tshedor:
@jamesmk If you can come up with a better solution than using the helpers (or even a solution that _works_ with the helpers, as the one on this branch doesn't), we should go with that. But I'm hitting a brick wall with the Ruby/Slim-only solution.

I guess I propose that you take a crack at it. I'm sorry; I have failed you.

tshedor:
Also, resolved the merge conflicts from your nav branch.

jamesmk:
@tshedor I think this looks good, very similar to your idea. LMK what you think.

spec/dummy/db/schema.rb
```diff
@@ -60,6 +60,14 @@
 
   add_index "coaches", ["team_id"], name: "index_coaches_on_team_id", using: :btree
 
+  create_table "dogs", force: :cascade do |t|
```
tshedor: @jamesmk did you mean to add this doge table?

jamesmk: Gah! Ok, fixed.

PR #91, opened 2016-04-13
log a console error when trying to create first user without admin role 

willchertoff:
I ran into this silent fail after dropping my db(post fae install). I couldn't figure out why Fae would not allow the creation of a new first user. After digging into the code, I found that fae fails silently if there isn't a 'super admin' role in the Fae::Role table (originally generated by fae:seed_db).  

Added a simple log message to indicate the error.

### DIFF
--- a/app/controllers/fae/setup_controller.rb
+++ b/app/controllers/fae/setup_controller.rb
@@ -1,6 +1,7 @@
 module Fae
   class SetupController < ActionController::Base
 
+    before_action :check_roles
     layout 'devise'
 
     def first_user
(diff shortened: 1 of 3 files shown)


jamesmk:
Thanks for the PR @willchertoff! This is a great point, but I would take is a step further. Could you add this as a `before_action` in the `SetupController` that raises an exception? Without these roles Fae can't function properly, so we might as well stop the process in it's tracks and make it obvious what the issue is.

thanks!

willchertoff:
Good idea @jamesmk. I have updated the ```SetupController``` with a ```before_action``` to make the issue more obvious.

jamesmk:
@willchertoff can you take a crack at fixing the spec in this PR? Since we're assuming the super admin role is there, will have to create it manually.

Adding this should do it:
```
FactoryGirl.create(:fae_role, name: 'super admin')
```

thanks!

jamesmk:
Looks good, thanks @willchertoff!

PR #93, opened 2016-04-13
add delete_path option to fae_delete_button

tshedor:
some edge cases for the fae_delete_button have @parent defined in the controller and not as fae_parent on the model. This allows a custom delete_path to be passed.

Also, I removed some extra markup to make it more future-update-proof.


jamesmk:
@tshedor Do you think you could add a test for this option?

tshedor:
@jamesmk added

PR #95, opened 2016-04-14
55244 namespaced models

jamesmk:


docs/helpers.md
```diff
@@ -457,6 +457,16 @@ fae_avatar(current_user)
 #=> 'https://secure.gravatar.com/....'
 ```
 
+## fae_sort_id
+
+This method returns an string suitable for the row IDs on a sortable table (you can make a table sortable by adding the `js-sort-row` class to it).
```
tshedor: LDT - "returns a string"

jamesmk:
thanks @tshedor, updated!

PR #100, opened 2016-04-19
Added in the ability for the new path on nested table to accept params

nilesvm:
This is useful for separating groups of objects by a parameter without having to create/maintain a whole separate object for each group, for example gallery images for 2 types of purposes on the same parent object.


app/views/fae/shared/_shared_nested_table.html.slim
```diff
@@ -44,9 +44,9 @@ table class=table_class
     - else
       tr
         - if index
```
jamesmk: This conditional is no longer needed, hooray! Combine L46-49 into one line:

```
- td_link = link_to "add some", new_path, class: 'js-add-link'
```

spec/dummy/app/views/admin/wines/_form.html.slim
```diff
@@ -21,10 +21,30 @@
...
       parent_item: @item,
       cols: [:name, :winemaker_image, :table_image],
+      new_path: "/admin/winemakers/new?region_type=1",
       ordered: true,
-      edit_column: true
+      edit_column: true,
+      header: 'Oregon Winemaker',
+      title: 'Oregon Winemakers'
+
+- if params[:action] == 'edit'
```
jamesmk: Do you need to call this conditional again?

PR #105, opened 2016-04-22
57158 - Fae::Navigation#current_section no longer removes '/new' from…

jamesmk:
… the middle of the URL string

### DIFF
--- a/app/models/fae/navigation.rb
+++ b/app/models/fae/navigation.rb
@@ -20,6 +20,10 @@ module Fae
       find_items_by_text(structure, query, [])
     end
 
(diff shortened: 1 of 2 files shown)


app/models/fae/navigation.rb
```diff
@@ -20,6 +20,10 @@ def search(query)
       find_items_by_text(structure, query, [])
     end
 
+    def current_section
+      @current_path.gsub(/\/new$|\/\d+\/edit/, '')
```
tshedor: Ohhhhhhhhhhhhhhhhh

PR #117, opened 2016-05-10
add Fae::NestedBaseController

jamesmk:
This adds a `Fae::NestedBaseController` in which future nested object controllers will inherit from. This resolves something that has been bugging me for a while: the ugly nested controller code that is generated in the app.

Looking for feedback from @tshedor @LeChin and @nilesvm


app/controllers/fae/base_controller.rb
```diff
@@ -78,12 +78,12 @@ def set_class_variables(class_name = nil)
...
     end
 
-    # Use callbacks to share common setup or constraints between actions.
+    # use callbacks to share common setup or constraints between actions.
     def set_item
       @item = @klass.find(params[:id])
     end
 
-    # Only allow a trusted parameter "white list" through.
+    # only allow a trusted parameters, override to whitelist
```
tshedor: LDT - if you grammatizing this, "only allow a trusted parameters"

docs/usage.md
```diff
@@ -790,3 +793,41 @@ def nav_items
...
+
+---
+
+# Nested Forms
+
+![Nested Table](images/nested_table.png)
+
+Fae provides an easy way to manage associated objects in one form via nested forms. A good use case for nested forms is when you have a `has_many` association where the child object is simple enough that an embedded table and form on the parent form will do. This makes managing the associated content much more convenient.
+
+Some good examples ideal for nested tables are: image galleries, tasting notes on a wine release and quotes on a person object.
```
tshedor: No colon needed after `are` OR remove `are` and put it after `tables`

docs/usage.md
```diff
@@ -790,3 +793,41 @@ def nav_items
...
+
+## Setting Up a Nested Form
+
+### Generate the child object
+
+Assuming the parent object already exists, generate the object to be nested with [Fae's nested scaffold command](#faenested_scaffold).
```
tshedor: `#faenested_scaffold` needs a space before `nested`

lib/generators/fae/templates/controllers/nested_scaffold_controller.rb
```diff
@@ -1,80 +1,5 @@
 module <%= options.namespace.capitalize %>
-  class <%= class_name.pluralize %>Controller < Fae::ApplicationController
```
tshedor: Yaaaaaaasssssssss

spec/dummy/app/controllers/admin/aromas_controller.rb
```diff
@@ -1,65 +1,8 @@
...
-      @parent_item = @item.release
-
-      if @item.destroy
-        flash[:notice] = 'Item successfully removed.'
-      else
-        flash[:alert] = 'There was a problem removing your item.'
-      end
-      render template: 'admin/aromas/table'
-    end
+  class AromasController < Fae::NestedBaseController
```
LeChin: 😻

LeChin:
Besides the code climate errors -- which I don't really know what a good fix would be, since that sorta stuff is used all over and it doesn't seem to complain about it -- look's great! Nice n clean.

docs/usage.md
jamesmk: That's the actual anchor tag: https://github.com/wearefine/fae/blob/master/docs/usage.md#faenested_scaffold

lib/generators/fae/templates/controllers/nested_scaffold_controller.rb
jamesmk: https://thepawsomelion.files.wordpress.com/2015/11/tumblr_lpwtgvjbym1qko4x4o1_500.gif

app/controllers/fae/nested_base_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+        flash[:alert] = t('fae.delete_error')
+      end
+      render template: table_template_path
+    end
+
+    private
+
+    def set_class_variables
+      @klass_name = params[:controller].split('/').last
+      @klass = @klass_name.classify.constantize
```
tshedor: This could be cause for concern. I don't have a solution, but this allows access to any model so long as you configure the URL to have the unclassified version. But I also recall us checking for roles when updating models through AJAX (like users couldn't delete users, only admins could)...and maybe that's the answer? I.e. it looks unsafe here, but it's harmless because there's that verification when a row is updated.

jamesmk: It's not ideal and I want to refactor these class instance variables in 2.0 anyway, but I don't consider it vulnerable. First these instance variables are only created on defined Fae controllers and used in the app in the same way a normal class would be used. You can't arbitrarily access this as a user, you'd hit a 404 or 401 first. Even if you could, i don't see any way to remotely execute on the instance variable.

If you see a specific method in which this is vulnerable, let me know.

tshedor: Alright, I'm good. @nilesvm did you have words to lend? Otherwise Imma murge

nilesvm: @tshedor all and all this update makes the code a bit cleaner and clearer. Merge it!

spec/dummy/app/controllers/admin/aromas_controller.rb
```diff
@@ -1,65 +1,8 @@
 module Admin
```
nilesvm: Look at all this red ~!

PR #118, opened 2016-05-21
57075 - add a --fine flag to the install generator

jamesmk:
Adds a FINE flage for internal usage. That way we can have a set of internal standards without making them global.

```
rails g fae:install --fine
```


tshedor:
Aw hell yeah

PR #123, opened 2016-06-10
Authorization

jasonfine:
@jamesmk let's look at the layer that stops the items from getting drawn, currently it's in the views. you wanted to avoid that but i am having trouble getting it to jive trying to do something in navigation.rb where the hash gets built.

another consideration is that the global search will pull results for models that the user may technically be off limits from accessing - however they will only be able to see them in the results, trying to click through to them will hit the controller-level access control and 404 for them.

thanks


jamesmk:
@jasonfine I've moved authorization of nav items to `navigation.rb` and removed the need to define anything in the navigation concern. 

I still want to review the naming and usage to make sure it's consistent with the other features and adds some documentation. We're close!

jamesmk:
@jasonfine @tshedor I think this PR is ready for some final eyes

docs/features/authorization.md
```diff
@@ -0,0 +1,63 @@
...
+          'content_blocks/about_us' => ['super admin', 'admin', 'custom role']
+        }
+      end
+
+    end
+  end
+end
+```
+
+## Custom Roles
```
tshedor: Do custom roles inherit from user or admin or super admin? Is it possible to define what permissions they inherit?

app/models/fae/navigation.rb
```diff
@@ -69,9 +72,28 @@ def item(text, options={})
       # prefer `subitems` over `sublinks` for new DSL but still need to support old structure in v1
       hash[:sublinks] = hash[:subitems] if hash[:subitems].present?
 
+      # authorize link with current user
+      unless can_view_path(hash[:nested_path])
+        hash[:path] = '#'
```
tshedor: So only the path is hidden from unauth'd users? I think this is unnecessary - I'd rather click and get an unauthorized / 404 page than to click and think the link is broken. Actually, I'd vote to remove the nav item entirely if the user isn't authorized.

jamesmk: The item is removed in the `recursive_authorization` method. This will flag the item for removal but also makes sure `Fae::Navigation#strucrture` never contains a path the `current_user` can't view.

docs/features/authorization.md
tshedor: I think that comment ^ is something to note, verbatim, in the docs

app/models/fae/navigation.rb
tshedor: Ah, I see. My bad for not seeing the flow of everything. LDT, but I think it'd be worth noting in the `navigation_concern.rb` docs that any `href="#"` without sublinks will be removed from nav items, because that isn't expected behavior. And I can see some case arising where the nav hosts some custom JS functionality using faux `href="#"` links.

jamesmk:
@tshedor updated docs with your suggestions

jamesmk:
testing slack hook

jamesmk:
@tshedor updated docs with your suggestions

tshedor:
@jasonfine @jamesmk  Seal of Approval

PR #124, opened 2016-06-10
57288 - show nested path in main nav

jamesmk:
This updates the main nav to always link to the first item in the dropdown by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
     + Generated nested table controllers are now empty and inherit from `Fae::NestedBaseController`
     + \#57075: Add a flag to the install generator for internal usage
     + Support multiple languages on static page fields
+    + \#57288: main nav items will display the paths of the first drop down item by default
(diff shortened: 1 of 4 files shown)


tshedor:
In the dummy app instance, this fix is sensible because the parent heading is not an actual page. But in most Fae installs, the top level can link to an index page, and this PR will require that the first sub level mirror the top level. For example:

```
L1 Movies (/admin/movies)
  L2 Actors (/admin/actors)
  L2 Genres (/admin/genres)
```

With the proposed PR, the top level path is unchangeable (i.e. `nested_path` isn't falling back to `path` if `path` is provided), so this is required:

```
L1 Movies
  L2 Films (/admin/movies)
  L2 Actors (/admin/actors)
  L2 Genres (/admin/genres)
```

I think `path` should be an allowable fallback if we move forward with this PR.

tshedor:
Nevermind, I was wrong about this.

PR #130, opened 2016-06-21
conditionally handle the data response from remotipart...

kyleshevlin:
... to ensure proper parsing.

The data in the response was returning an array: [ <style></style>, <textarea/> ] and thus line 124 would pass but 125 would fail. This would prevent the rest of the functionality from firing.

By checking if the response had a length greater than 1, I was able to loop to find the textarea object and then parse it to send the proper data for rendering to the page.

### DIFF
--- a/app/assets/javascripts/fae/form/_ajax.js
+++ b/app/assets/javascripts/fae/form/_ajax.js
@@ -143,7 +143,9 @@ Fae.form.ajax = {
             // $textarea just contains a string, this will parse it as HTML and turn it into a jQuery object
             // http://stackoverflow.com/a/16859718/2347675
             if ( textarea ) {
-              $html = $( $.parseHTML(textarea.text()) );
+              debugger;
+              $html = $( textarea.text() );
+              debugger;
             }
           } else {
             $html = $( $html.val() ); // Original handling
--- a/app/assets/javascripts/fae/form/_ajax.js
+++ b/app/assets/javascripts/fae/form/_ajax.js
@@ -124,7 +124,7 @@ Fae.form.ajax = {
           if ( $html.length > 1 ) {
             var textarea;
 
-            function isTextarea(items) {
+            var isTextarea = function(items) {
               for (var i = 0; i < items.length; i++ ) {
                 var $item = $( items[i] );
 
--- a/app/assets/javascripts/fae/form/_ajax.js
+++ b/app/assets/javascripts/fae/form/_ajax.js
@@ -143,9 +143,7 @@ Fae.form.ajax = {
             // $textarea just contains a string, this will parse it as HTML and turn it into a jQuery object
             // http://stackoverflow.com/a/16859718/2347675
             if ( textarea ) {
-              debugger;
               $html = $( textarea.text() );
-              debugger;
             }
           } else {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/fae/form/_ajax.js
```diff
@@ -121,7 +121,21 @@ Fae.form.ajax = {
 
         // remotipart returns data inside textarea, let's grab it from there
         if ($html.is('textarea')) {
-          $html = $( $html.val() );
+          if ( $html.length > 1 ) {
+            var $textarea;
+
+            for (var i = 0; i < $html.length; i++ ) {
+              if ( $($html[i]).is('textarea') ) {
```
tshedor: Something seems wrong here. If, on L123, `$html === 'textarea'` then it should not be an Array. Does this code work in the dummy app?

kyleshevlin: Line 123 is not the same as ( $html === 'textarea'), is() responds true if any item in the array is a textarea.

tshedor: Interesting, did not know that. So then what if L128 returns another Array?

Other notes: if `$textarea` is not reassigned in this loop, L135 will throw an error (`.text()`) and there should be a changelog item added for this.

kyleshevlin: I'll tidy this up.

app/assets/javascripts/fae/form/_ajax.js
```diff
@@ -121,7 +121,33 @@ Fae.form.ajax = {
 
         // remotipart returns data inside textarea, let's grab it from there
         if ($html.is('textarea')) {
-          $html = $( $html.val() );
+          if ( $html.length > 1 ) {
+            var textarea;
+
+            var isTextarea = function(items) {
```
tshedor: 1. A function like this should be defined outside of this `if` block, like above the `ajax:success`. 
2. `textarea` can be assigned on creation.  (L125 and L141)
3. Executing HTML of a textarea seems unsafe (L146). Is `$( textarea.text() )` insufficient? 
4. If this original problem stems from `.is('textarea')` being an Array, should L123 check for this to be `$.isArray()`? And that should remove the `if` on L124 and the need for the extra nesting on L145-50.

tshedor: Also, changelog item plz

kyleshevlin: textarea.text() is not a jQuery object which we still need for the rest of the function, $.parseHTML will return a jQuery object.

I didn't want to remove any previously existing functionality because it seemed like I was dealing with a strange case, so I kept the check for .is() in there with the response it got originally in the else of the nested if/else block. The first .is('textarea') will determine if there is any textarea present, then we can recursively check for the exact textarea; if there is no textarea, skip it. We recursively check because as you pointed out, while unlikely, perhaps the textarea is nested another level deep. I think it's overkill, but I was addressing your concern above.

jamesmk:
@kyleshevlin Since we can't consistently see this issue I'm just going to close this. Sorry for the goose chase, if other AdBlock user start seeing this I'll pluck these commits.

PR #131, opened 2016-06-30
55333 - Pagination (with filtering and column sorting refactor)

jamesmk:
As I went to add the ability to paginate, I quickly realized I would need to incorporate it into the filtering method or the two wouldn't play nice together. To get them to talk to each other I added deeplinks (via Fryr) to both.

Then I realized our current column sorting doesn't work with pagination because it only sorts the visible items in the table and that needed a second approach. This PR incorporates those three features into one mechanism.

I have some TODOs still, but wanted to get some eyes on it as soon as I could. Specifically the changes in developer interface. This adds a second way to sort columns, which I don't like, but the original way is so much easier for tables without pagination. I thought it made sense to keep it.

@tshedor: I could use some help on adding the sorting arrows to the new sorting method, we can chat IRT.

@LeChin: I'm pretty sure this breaks filter cookies, but I didn't see a use case in the dummy. I may need some help on that too.

thanks!


LeChin:
Hokay, I'll take a look at this later today/tomorrow. Want me to implement some filter cookies in the dummy and push to dev? @jamesmk

jamesmk:
Thanks @tshedor and @LeChin! I'll add some more tests and docs over the weekend and I think it's about ready. If you see anything else to fix, add a comment and I'll get to that too.

tshedor:
Thanks for adding the documentation. @LeChin I'm cool with this.

PR #137, opened 2016-07-28
refs #59386 - show sub page indicator as "open" when landing on an ac…

tshedor:
…tive view


jamesmk:
@tshedor oh em gee, conflicts!

PR #140, opened 2016-08-02
Page with lang fix

jamesmk:
Since the pages associated are only defined the first time they are called, the default association on static page attributes should default to something more predictable.

Currently the default association is set to whatever the locale of the person to call it has set and will stay that way until restart.

cc @tshedor

### DIFF
--- a/app/models/fae/static_page.rb
+++ b/app/models/fae/static_page.rb
@@ -39,7 +39,7 @@ module Fae
             define_validations("#{name}_#{lang}", type, value[:validates]) if value.try(:[], :validates).present?
           end
           # Save with lookup to have default language return in front-end use (don't need to worry about validations here)
-          define_association(name, type, "#{name}_#{I18n.locale}")
+          define_association(name, type, "#{name}_#{languages.first}")
         else
           # Normal content_blocks
           define_association(name, type)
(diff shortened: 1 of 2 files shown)


app/models/fae/static_page.rb
```diff
@@ -39,7 +39,7 @@ def self.setup_dynamic_singleton
             define_validations("#{name}_#{lang}", type, value[:validates]) if value.try(:[], :validates).present?
           end
           # Save with lookup to have default language return in front-end use (don't need to worry about validations here)
-          define_association(name, type, "#{name}_#{I18n.locale}")
+          define_association(name, type, "#{name}_#{languages.first}")
```
tshedor: Instead of `languages.first` we should use the Rails default locale setting, which is more intuitive: `Rails.application.config.i18n.default_locale`

tshedor:
Correct me if I'm wrong, but this still doesn't solve for the issue we're experiencing on the project. According to @LeChin, a helper is required for rendering international static page content on the front-end, right? That helper should be added to the internationalization docs.

app/models/fae/static_page.rb
jamesmk: `Rails.application.config.i18n.default_locale` can be `nil`, so I'll still use `languages.first` as a fallback.

jamesmk:
@tshedor You're correct in that this doesn't solve for @LeChin. Ultimately the issue has to do with how the FE selects which language attr to use, and I don't know how much of that logic I want to bake into Fae. Fae's language helpers should be more about making the management of the content easier and not be a full stack i18n solution.

PR #142, opened 2016-08-10
Use pagination

nilesvm:
Basically, only 25 items are displaying on index views because of the recently added pagination feature #55333.

If FAE is at the head of the unreleased version, and a dev has not built an index page w pagination scaffolding, only the default number of results are displaying and the rest are hidden. So if there are 75 items, and the default is 25, only 25 are displaying.

### DIFF
--- a/docs/features/filtering.md
+++ b/docs/features/filtering.md
@@ -211,7 +211,7 @@ Under the hood Fae uses [Kaminari](https://github.com/amatsuda/kaminari) to powe
 
 ### Enable Pagination in the Controller
 
-Pagination is disabled in the Fad:BaseController by default. This can be customized with a private use_pagination method in controllers inheriting from Fae::BaseController.
+Pagination is disabled in the Fae:BaseController by default. This can be customized with a private use_pagination method in controllers inheriting from Fae::BaseController.
 
 ```ruby
 # app/controllers/admin/example_controller.rb
--- a/docs/features/filtering.md
+++ b/docs/features/filtering.md
@@ -211,7 +211,7 @@ Under the hood Fae uses [Kaminari](https://github.com/amatsuda/kaminari) to powe
 
 ### Enable Pagination in the Controller
 
-Pagination is disabled in the Fae:BaseController by default. This can be customized with a private use_pagination method in controllers inheriting from Fae::BaseController.
+Pagination is disabled in the Fae:BaseController by default. This can be customized with a private `use_pagination`method in controllers inheriting from Fae::BaseController.
 
 ```ruby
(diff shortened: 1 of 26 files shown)


jamesmk:
@nilesvm Thanks for jumping on this! This is a good start, unfortunately this has to be a per controller configuration. Some sites would use pagination only in a couple sections, so turning this on would see the same issue in the non-paginated sections.

nilesvm:
Well, you could just over-ride the index method on the applications controller if you wanted to do it per controller. 

Alternatively, perhaps we could add in a before_action method to base controller w a default of false. On controllers in the app where we wanted to add pagination we could just update this method to be true.

`
    def use_pagination
      false
    end
`

` 
def index
      if Fae.use_pagination || use_pagination
        @items = @klass.for_fae_index.page(params[:page])
      else
        @items = @klass.for_fae_index.all
      end....
`

What do ya think haus?

jamesmk:
I think per controller is our main use case, I don't think we've have an app that used pagination across the board. In that case I'd prefer to keep the config in one place and the controller makes the most sense to me. If you need this as a stop gap solution, go for it and we can address next week.

thanks

docs/features/filtering.md
```diff
@@ -209,6 +209,27 @@ Under the hood Fae uses [Kaminari](https://github.com/amatsuda/kaminari) to powe
 * ignore `fae_paginate` and use Kaminari's `paginate` directly
 * etc...
 
+### Enable Pagination in the Controller
+
+Pagination is disabled in the Fad:BaseController by default. This can be customized with a private use_pagination method in controllers inheriting from Fae::BaseController.
```
LeChin: what's this Fad app you speak of? Also, can you put ticks around the `use_pagination` method?

spec/dummy/db/schema.rb
```diff
@@ -111,7 +111,8 @@
     t.boolean  "required",                  default: false
   end
 
-  add_index "fae_files", ["fileable_id", "fileable_type"], name: "index_fae_files_on_fileable_id_and_fileable_type", using: :btree
```
LeChin: where did these changes come from?

docs/features/filtering.md
nilesvm: you got it!

spec/dummy/db/schema.rb
nilesvm: I had to clean out and run a migration on the database from scratch. I believe that the changes  are just the order of where they appear in the schema.

LeChin: yea we good

PR #144, opened 2016-08-15
57064 custom trackable kids

jasonfine:

PR #145, opened 2016-08-16
lori wants the table to fade while filtering is loading because for m…

jasonfine:
…ore intense logic it appears that nothing is happening or something is borked to the user

### DIFF
--- a/app/assets/javascripts/fae/form/_filtering.js
+++ b/app/assets/javascripts/fae/form/_filtering.js
@@ -31,6 +31,7 @@ Fae.form.filtering = {
     // hardcode _this because this === Fryr object
     var _this = Fae.form.filtering;
     var post_url = _this.$filter_form.attr('action');
+    $('.js-sort-row').css({opacity: 0.5});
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/fae/form/_filtering.js
```diff
@@ -31,6 +31,7 @@ Fae.form.filtering = {
     // hardcode _this because this === Fryr object
     var _this = Fae.form.filtering;
     var post_url = _this.$filter_form.attr('action');
+    $('.js-sort-row').css({opacity: 0.5});
```
jamesmk: @jasonfine I'd rather not apply CSS with JS. Can you apply a class with JS and use CSS on the class?

thanks!

jamesmk:
@jasonfine Can you add a changelog item for this as well?

jasonfine:
all set

jamesmk:
@jasonfine Oops, conflict from #144. Fix that up and we're good to go!

PR #149, opened 2016-09-19
57081 titleize

LeChin:
Change labels, ths, button text and other relevant areas to .titleize instead of .humanize.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,3 @@ spec/dummy/public/uploads/*
 .rubocop.yml
 
 *.gem
-
-vendor/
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -24,3 +24,5 @@ spec/dummy/public/uploads/*
 .rubocop.yml
 
 *.gem
+
+vendor/
```
jamesmk: @LeChin same for here too, please remove this from the gitgnore

jamesmk:
@LeChin I'm seeing a bunch of failing tests around the use of `titleize`. Can you address those?

thanks!

LeChin:
@jamesmk removed item from gitignore and fix failing tests due to typo

PR #160, opened 2016-10-04
61163 fix sortable stick columns

aaronransley:


app/assets/javascripts/fae/form/_filtering.js
```diff
@@ -179,7 +183,7 @@ Fae.form.filtering = {
   _addSortingListeners: function() {
     var _this = this;
 
-    $('th[data-sort]').on('click', function() {
+    $('body').on('click', 'th[data-sort]', function() {
```
tshedor: 👋 Use `FCH.$body`

app/assets/javascripts/fae/form/_filtering.js
```diff
@@ -152,7 +152,11 @@ Fae.form.filtering = {
    */
   sortingSetup: function() {
     this._addSortingClasses();
-    this._addSortingListeners();
+
+    if (!this._are_sorting_listeners_setup) {
+      this._addSortingListeners();
+      this._are_sorting_listeners_setup = true;
```
tshedor: Personal preference, but I'd elect to include this at the end of `_addSortingListeners()`. That way, if the variable is mutated again, it's always reset after this function is called. Otherwise you'll have to pray that whenever the function is invoked someone will always declare `this._are_sorting_listeners_setup` immediately afterward.

app/assets/javascripts/fae/_tables.js
```diff
@@ -294,6 +297,10 @@ Fae.tables = {
 
       $fixed_header.append( $header );
       $this.after($fixed_header);
+
+      if ($this.hasClass('tablesorter')) {
+        _this.proxyStickyTableHeaderClicks($fixed_header, $this);
```
tshedor: This function, as well as the newly added `replaceStickyTableHeader` are only invoked in one location each in this file. Are they being called outside this file? I'm pro-modularization, but these both seem like unnecessary public functions. 

I don't think you'd be wrong to include the contents of these functions instead here, since they're only 1 -5 lines. The added `_this` can also be removed.

app/assets/javascripts/fae/form/_filtering.js
LeChin: HALLO TIM

tshedor: Guten tag

app/assets/javascripts/fae/form/_filtering.js
aaronransley: Good call. Looking back at it, I'll try moving the check for `this._are_sorting_listeners_setup` into `_addSortingListeners()` and have it trigger an early return if set.

app/assets/javascripts/fae/form/_filtering.js
aaronransley: @tshedor It looks like FCH.$body returns an empty array. Any idea why that might be happening?

app/assets/javascripts/fae/form/_filtering.js
```diff
@@ -179,7 +179,10 @@ Fae.form.filtering = {
   _addSortingListeners: function() {
     var _this = this;
 
-    $('th[data-sort]').on('click', function() {
+    // Fizzle if listeneres have already been attached
+    if (this._are_sorting_listeners_setup === true) { return; }
+
+    $('body').on('click', 'th[data-sort]', function() {
```
aaronransley: Continuing to use `jQuery('body')` until issue of `FCH.$body` returning `[]` is resolved

aaronransley: Continuing to use `$('body')` until issue of `FCH.$body` returning `[]` is resolved

app/assets/javascripts/fae/_tables.js
```diff
@@ -72,6 +72,13 @@ Fae.tables = {
 
         // Save it to the cookie as a string
         Cookies.set(_this.sort_cookie_name, cookie_value);
+
+        // Replace sticky table headers w/ newly sorted headers
+        $this.siblings('.sticky-table-header')
```
aaronransley: Sticky table header replacement now inline

app/assets/javascripts/fae/form/_filtering.js
tshedor: @aaronransley that is unusual. Educated guess is that Frob Core is being initialized before the DOM is ready, and therefore jQuery returns an empty array since it can't find the selector. There doesn't appear to be `FCH.$body` used elsewhere in the code, so ignore my comments. 

This should be an issue on [Frob Core](https://github.com/wearefine/frob-core). I'm surprised it hasn't come up in other projects.

tshedor:
Can you approve a whole review? Is that a thing on GitHub? I approve the whole PR.

jamesmk:
@tshedor you can start a code review with your comments, then once they're address mark it as approved. Still getting used to the feature.

PR #164, opened 2016-10-14
Content form html

nilesvm:
Fixed bug that prevented content_form from displaying custom input_html.


jamesmk:
@nilesvm: ❤️ this PR, use case in dummy and tests too!

PR #171, opened 2016-10-28
Refs #60104 - Auto expand collapsible tables on page load

aaronransley:
Issue: https://issues.afinedevelopment.com/issues/60104

### DIFF
--- a/app/assets/javascripts/fae/_tables.js
+++ b/app/assets/javascripts/fae/_tables.js
@@ -267,6 +267,8 @@ Fae.tables = {
 
       Fae.tables.sizeFixedHeader();
     });
+
+    $toggle.trigger('click');
   },
 
   /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - enhancements
 - bugs
+    + \#60104: Default to expanded collapsible tables on page load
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/fae/_tables.js
```diff
@@ -267,6 +267,8 @@ Fae.tables = {
 
       Fae.tables.sizeFixedHeader();
     });
+
+    $toggle.trigger('click');
```
jamesmk: @aaronransley Did you look into updating classes and CSS first? I would prefer rendering the open-by-default state to begin with via CSS vs rendering them closed and using a JS click to open them. What do you think?

aaronransley: I did, but discovered that there is also static markup in admin views outputting "Open All" on page load. Looks to be spread across Fae installs and falls outside an easily modifiable helper method. Simulating a click seemed to be the lowest touch way to handle this ask, but I agree it feels a bit tacked on 🤔

tshedor: I agree with James. It's adding cruft; instead, the `.active` should be added to `.collapsible` in the dummy app and the docs (and the Open All changed to Close All and `.close-all` added to `.collapsible-toggle`). Simulating the click will have unintended consequences later.

PR #174, opened 2016-11-11
adds support for standard_table

markhoffmann:
retains old traversal code for backwards compatibility

### DIFF
--- a/app/assets/javascripts/fae/form/_filtering.js
+++ b/app/assets/javascripts/fae/form/_filtering.js
@@ -37,7 +37,12 @@ Fae.form.filtering = {
       var $data = $(data);
       var $table_from_data = $data.find('table').first();
       // replace table
-      _this.$filter_form.next('table').replaceWith($table_from_data);
+      if ($(".standard_table").length > 0) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/fae/form/_filtering.js
```diff
@@ -37,7 +37,12 @@ Fae.form.filtering = {
       var $data = $(data);
       var $table_from_data = $data.find('table').first();
       // replace table
-      _this.$filter_form.next('table').replaceWith($table_from_data);
+      if ($(".standard_table").length > 0) {
```
tshedor: Is `.standard_table` a class in another app? I don't see it in the `v1.4.1` branch. This will be confusing in the future to maintain if it's specifically for an unrelated app.

Instead, I'd name it something more specific and indicative of what you're trying to do, i.e. `.js-ajax-table-data` (`.js-` prefix since it's unrelated to CSS).

The `if` should also be condensed because if `.length` returns 0 it's falsey and single quotes are preferred.

tshedor: And it should be documented, changlog, yada yada

PR #175, opened 2016-11-11
60183-fix adjust modal styles for images so smaller images aren't bei…

JuliannaGreen:
…ng squeezed into non-existence by too much padding

### DIFF
--- a/app/assets/stylesheets/fae/modules/_modal.scss
+++ b/app/assets/stylesheets/fae/modules/_modal.scss
@@ -22,7 +22,6 @@
   position: relative;
   background: $c-modal-bg;
   color: $c-darker-grey;
-  padding: 20px;
   line-height: 1.5;
   font-size: 13px;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/fae/modules/_modal.scss
```diff
@@ -56,16 +55,20 @@
...
     max-width: 100%;
     max-height: 100%;
+    margin: auto;
+    display: block;
   }
 }
 
 .simplemodal-close {
+  display: block;
+  height: 30px;
```
tshedor: Locking in the height and appending `55` to the dimensions in JS may not be the best solution here. Does using `box-sizing: content-box;` work? I'd prefer to use CSS for anything that doesn't dynamically calculate the size of an object.

jamesmk: I'm going to merge this change. This is update is 🔥 so we can look into ways to refactor later.

PR #180, opened 2016-12-13
:fire: fix filtering bug

LeChin:


app/assets/javascripts/fae/form/_filtering.js
```diff
@@ -31,17 +31,17 @@ Fae.form.filtering = {
     // hardcode _this because this === Fryr object
     var _this = Fae.form.filtering;
     var post_url = _this.$filter_form.attr('action');
-    var $results_table = ($(".js-results-table").length) ? $(".js-results-table").first() : _this.$filter_form.find('table');
- 
+    var $results_table = ($(".js-results-table").length) ? $(".js-results-table").first() : _this.$filter_form.next('table');
```
tshedor: This won't break older 1.x versions of Fae (pronounced "yae"), right? Because this call is looking after the form, not within it. If this does break older versions, an ugly hack will cover that edge case, inserted after this line.
```javascript
// @depreciation - remove this control in version 2.0 and adjust above ternary to be `.next` not `.find`
if (!$results_table.length) {
  $results_table = _this.$filter_form.next('table');
}
```

LeChin: I can't speak to that. From my understanding, the PR that introduced this code was heavily vetted, so I ASSume it should be fine...

jamesmk: @tshedor This is fixing an update that broke old implementations: https://github.com/wearefine/fae/pull/174/files#diff-cebcdb529c6537915f3dae2fe8212a79L40

tshedor: @jamesmk @LeChin Never change, guys. Thanks.

PR #181, opened 2016-12-13
hotfix for pagination on filter form

LeChin:
Not sure if this is too clever, but I didn't want a ton of nested if/elses to account for use_pagination method _and_ reset search - need this for same project as last bug fix was for.

credit: http://stackoverflow.com/a/1797235/5261969

### DIFF
--- a/app/controllers/fae/base_controller.rb
+++ b/app/controllers/fae/base_controller.rb
@@ -62,11 +62,12 @@ module Fae
     end
 
     def filter
-      if params[:commit] == "Reset Search"
-        @items = @klass.filter_all.page(params[:page])
-      else
-        @items = @klass.filter(params).fae_sort(params).page(params[:page])
-      end
+      return_items = []
+      return_items << [:page, params[:page]] if use_pagination
+      return_items << [:filter_all] if params[:commit] == 'Reset Search'
+      return_items << [:filter, params] if params[:commit] != 'Reset Search'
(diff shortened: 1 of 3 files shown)


app/controllers/fae/base_controller.rb
```diff
@@ -62,11 +62,12 @@ def destroy
...
-        @items = @klass.filter_all.page(params[:page])
-      else
-        @items = @klass.filter(params).fae_sort(params).page(params[:page])
-      end
+      return_items = []
+      return_items << [:page, params[:page]] if use_pagination
+      return_items << [:filter_all] if params[:commit] == 'Reset Search'
+      return_items << [:filter, params] if params[:commit] != 'Reset Search'
+
+      @items = return_items.inject(@klass) { |a, e| a.send(*e) }
```
LeChin: note: a = object, e = method and arguments.. i can rename to be easier to read, rubocop was complaining about it.

jamesmk: I would prefer descriptive arguments over avoiding line length violations

jamesmk: @LeChin The readability of the update is still rubbing me the wrong way. Is the issue that `page` is undefined on certain objects? Maybe we could use `try` here:

```
@items = @klass.filter_all.try(page, params[:page])
```

PR #182, opened 2016-12-14
Refs #62471: Form accessibility improvements

aaronransley:
* Focus styles on Chosen instances, checkboxes, radio button groups, two-pane multiselect instances, and file upload buttons
* Improved keyboard navigability + interaction within forms
  * Allow tabbing through Markdown / SimpleMDE instances
  * Allow shift-tabbing from two-pane multiselect instances
  * Support for checkbox interaction using spacebar

I'm ok with the CSS changes, but have some concerns around the additions to `app/assets/javascripts/fae/form/_ajax.js`. There are a few `@todo` and `@hack` tokens sprinkled throughout the comments, but to summarize:

1. Some of the event listeners attached to `$('#js-main-content')` could be moved to their own namespaces + files. For example, checkbox events in the existing `app/assets/javascripts/fae/form/inputs/_checkbox.js` file; others in new files? I'm not sure why the existing events were bound here within the `_ajax.js` file, but given time restraints, I opted to add additional handlers here rather than refactor.
1. The DOM traversal techniques used in the `.ms-list keydown` handler seem to work well within the Fae dummy app, but my worry is that they make break down in the future. I was able to test various combinations of form elements in different orderings, but if new elements are added and somehow fail to adhere to the class naming conventions...?
    * In terms of addressing this, I wonder if a kind of "form model" could be created on page ready and be used to determine form tab-order and changes in focus between elements? Though it'd probably have to be reconstructed on DOM mutation as well... Thoughts?

Words.


app/assets/javascripts/fae/vendor/jquery.multi-select.js
```diff
@@ -221,10 +221,12 @@
               tabindex = (e.shiftKey) ? tabindex-1 : tabindex+1;
               $('[tabindex="'+(tabindex)+'"]').focus();
               return;
-            }else{
-              if(e.shiftKey){
-                that.$element.trigger('focus');
-              }
+            } else {
```
tshedor: It'd be better to remove this since it's already committed to the git history.

app/assets/javascripts/fae/form/_ajax.js
```diff
@@ -302,9 +307,69 @@ Fae.form.ajax = {
...
+      })
+
+      /**
+       * Support for shift+tab off of ms-list element.
+       * By default, $.multiSelect() plugin captures shift+tab and disgards it
+       * @todo This entire method feels very brittle. Possible alternative:
+       * - Create an index of all focusable form elements on page load / DOM mutation
+       * - Use this index to navigate upwards from .ms-list element
+      */
+      .on('keydown', '.ms-list', function(e) {
```
tshedor: Instead of sniffing out the previous input through traversal, could you store an input once it's focused as the last focused input? You'd have to be careful to do this per form (i.e. `$('input').focus(function() { $(this).closest('form').data('last-focused', $(this)); }`) so as to not run afoul of pages with multiple forms such as the AJAX'd ones. 

Once you have that, you could access the last input: `var $prevFocusable = $(this).closest('form').data('last-focused');` and do the rest of the `.focus()` and `.trigger` work.

tshedor: Also, be sure use three `===` instead of two `==`. L339 and L355

tshedor:
To answer your first point, they're in `_ajax.js` because they need to be reinitialized when new nodes are inserted. However, since they're bound to the container, this doesn't make much sense. I like the idea of moving them to the checkbox file, and would encourage it when time permits.

app/assets/javascripts/fae/vendor/jquery.multi-select.js
jamesmk: I'm not a big fan of hacking libraries like this. If the time comes to upgrade, this type of thing could easily get lost in the mix. However, I don't have a simple alternative.

PR #189, opened 2017-02-14
60892 add has_image and has_file methods per specs in ticket

jasonfine:


jasonfine:
this + 60891 will require some pretty heavy documentation updates

PR #190, opened 2017-02-14
60891 attachment generators

jasonfine:
this depends on 60892 getting pulled in first


spec/dummy/config/routes.rb
```diff
@@ -2,6 +2,49 @@
   root 'pages#home'
 
   namespace :admin do
+    resources :jasaaas
```
jamesmk: @jasonfine I think these resources can go :)

jasonfine: oops, gone

PR #201, opened 2017-03-21
65539 deprecating has_top_nav

jamesmk:
Deprecating the option to omit the top nav and `has_top_nav`. It is difficult to maintain the two nav types when all production instances now use the top nav.

Any instances using no top nav will automatically convert to having a top nav if upgraded.


spec/dummy/config/initializers/fae.rb
```diff
@@ -12,6 +12,7 @@
     ja: 'Japanese'
   }
 
+  # has_top_nav has been deprecated, but keeping in to ensure definition doesn't rasie exceptions
```
tshedor: `raise`. Do you want to note what version it's officially disappearing from? `@depreciation - ... in vX` is how I did it in the JS/SCSS/Slim. Makes it easy to search for in the code before the version is released.

docs/upgrading/index.md
```diff
@@ -1,10 +1,15 @@
...
 
+* [To v1.5](#to-v15)
 * [From v1.2 to v1.3](#from-v10-to-v11)
 * [From v1.1 to v1.2](#from-v11-to-v12)
 
 ---
 
+# To v1.5
+
+`has_top_nav` has been deprecated. Any admin still using side nav only will convert to using the top nav upon upgrade without any changes necessary.
```
tshedor: Big fan.

spec/dummy/config/initializers/fae.rb
jamesmk: All of the `@deprecation`s have been noting when the feature will be deprecated to easily remove it come 2.0. I think the changelog is the best place to track features that have been deprecated.

PR #210, opened 2017-04-08
should not singularize page class name

lengshuiyulangcn:
if you run a page generation at the first time, for example, as in the document

```
rails g fae:page AboutUs hero_image:image headline:string body:text
```

you will get `AboutU` added to the ContentBlocksController instead of `AboutUs`


jamesmk:
Thanks for catching that @lengshuiyulangcn 👍

PR #207, opened 2017-03-26
#167 - remove depreciation warnings in rails 5

tshedor:
I don't know if this is going to work, I want to see what happens when the tests run. I'm having trouble getting the tests to run locally with Appraisals.

PR #220, opened 2017-04-14
Fix depreciation warning from parametrize method

lilfaf:
Note that this warning has been removed in Rails 5.1 but anyway we
better use keyword argument

See : https://github.com/rails/rails/commit/0189f4db6fe518de8909b66b7f30046bac52dedc


jamesmk:
@lilfaf Thanks for the PR! Unfortunately, this change is already duplicated in #207. We just need to verify backwards compatibility on that one.

PR #221, opened 2017-04-14
Use mini_magick instead of rmagick adapter for image processing

lilfaf:
RMagick is such a pain to intall because of native extensions failing to
compile with imagemagick v7.x installed.

MiniMagick is much better alternative, ligther dependency and far
sufficient for fae requirements.

You have such a great doc ! This project deserves an easy setup 😉


jamesmk:
@lilfaf I love this idea, our team is starting to feel the pain of ImageMagick 7. However, MiniMagick isn't a drop-in replacement for Rmagick, so I'll have to test backwards compatibility.

thanks

6temes:
It would be a good idea to update CarrierWave to the Version 1, as well.

PR #223, opened 2017-04-17
Fix typos in documentation and test

6temes:
Just a small typo.

`onject -> object`


jamesmk:
Thanks @6temes! Funny the same random typo existed in three places, at least we're consistent! 😉

6temes:
English is not my first language, so I had to double-check on Internet because, for a moment, I thought that _onject_ could be an actual term that I didn't know it existed!

PR #239, opened 2017-05-25
adds documentation for row sorting feature

y-chiasson:
Hi, I needed to implement row sorting feature, started implementing all the code myself, then I realized that Fae already implemented that using jQuery sortable and a utilities controller for handling the model updates.. so I wrote documentation for it.

Also, I couldn't figure out why row sorting wasn't getting initialized after adding the `js-sort-row` class to the table... It turns out there is this code in [app/assets/javascripts/fae/_tables.js#L29](https://github.com/wearefine/fae/blob/master/app/assets/javascripts/fae/_tables.js#L29)

```javascript
if (FCH.exists('.sortable-handle') || FCH.exists('.main_content-sortable-handle')) {
  this.rowSorting();
 }
```

which requires the `sortable-handle` class to exist on a `td` element.. so I documented that too.

### DIFF
--- a/docs/features/filtering.md
+++ b/docs/features/filtering.md
@@ -19,6 +19,7 @@ end
   - [Fae Filter Select](#fae-filter-select)
 * [Pagination](#pagination)
 * [Column Sorting](#column-sorting)
+* [Row Sorting](#row-sorting)
 
 ---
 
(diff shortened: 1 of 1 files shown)


jamesmk:
This is all generated for you if you use the `rails g fae:scaffold` with a position attr, but we are missing documentation on the feature. This is a good start, I'll merge this and expand on it.

Thanks for the PR @y-chiasson!

PR #247, opened 2017-06-12
refs #67773 - test suite love

andrewdang17:
**Test Speed**
Individual tests files I changed run faster but when running the entire test suite we maybe saved like 2 seconds...I used `build_stubbed` which is faster than `build/create`, combined some feature tests to avoid logging in and revisiting the same route, and dryed up some tests. Not sure what else can be done to speed these tests up (significantly).

**Test Coverage**
Coverage looks pretty good already, ~96% with SimpleCov. The other 4% is just some simple controller tests that I think would be unnecessary to add but I could add them if you want.

**Flickering Tests**
Didn't come across any


andrewdang17:
conflicts resolved

PR #250, opened 2017-06-20
Add message option for fae_delete_button

JuliannaGreen:


app/helpers/fae/view_helper.rb
```diff
@@ -56,10 +56,10 @@ def fae_clone_button(item)
       end
     end
 
-    def fae_delete_button(item, delete_path = nil, *custom_attrs)
+    def fae_delete_button(item, delete_path = nil, message=nil, *custom_attrs)
```
jamesmk: @JuliannaGreen This change isn't backwards compatible as previous usage of the `custom_attrs` will break. However, thinking of a solution I realized you might not have to change the source at all. Can you try...

```ruby
fae_delete_button item, nil, data: { confirm: 'Destroy this item COMPLETELY' }
```

jamesmk:
Closing this as the same thing is already possible with

```ruby
fae_delete_button item, nil, data: { confirm: 'Destroy this item COMPLETELY' }
```

PR #251, opened 2017-06-20
refs#67950-remove appended input label from validation errors

JuliannaGreen:
I noticed while I was testing that we are checking for presence while simultaneously allowing blank for validations in this file: https://github.com/wearefine/fae/blob/master/lib/fae/validation_helper_collection.rb

Should I fix and add to this PR? 

Also, do I need to add to changelog? If so, where?


jamesmk:
No changelog item needed. With community PRs coming in, I've been adding all changelog items before a release.

PR #267, opened 2017-07-13
refs 68776 - fix search form clearing

andrewdang17:


andrewdang17:
I couldn't simulate an pressing an enter key in capybara using [this] and some other workarounds (http://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FNode%2FElement%3Asend_keys) so I'm short one test that we discussed but I don't think it's worth spending more time on. What do you think?

PR #270, opened 2017-07-14
acronym inflection and helper text class

jasonfine:
the only one it doesn't work on is pH [they have to start with a capital letter] - i guess we can live with that.


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 ## 1.6
 
 - enhancements
+    + Add acronym inflections
```
jamesmk: You can omit the changelog items, I'm building them with each release now

lib/fae/helper_texter.rb
```diff
@@ -0,0 +1,20 @@
+class HelperTexter
```
jamesmk: The way you implemented this will only make use of it when using Fae generators. I'm wondering if a better spot is in `Fae::FormHelper` where you could call a private method from `fae_input` that sets these defaults? What do you think @jasonfine?

spec/dummy/config/initializers/inflections.rb
```diff
@@ -11,6 +11,13 @@
 # end
 
 # These inflection rules are supported but not enabled by default:
-# ActiveSupport::Inflector.inflections(:en) do |inflect|
-#   inflect.acronym 'RESTful'
-# end
+ActiveSupport::Inflector.inflections(:en) do |inflect|
```
jamesmk: You should add this to `config/initializers/inflections.rb` (outside of the dummy). Then it will be available to any app using Fae. Here, it'll only affect the dummy app.

jasonfine:
overhauled per suggestions

jasonfine:
hold on this, i see the failing tests. will get to them when i can

jasonfine:
tests fixed

PR #279, opened 2017-07-19
call unique_changeable_types on class not item set

jasonfine:


app/views/fae/pages/activity_log.html.slim
```diff
@@ -4,7 +4,7 @@ main.content
 
   == fae_filter_form title: 'Search Changes', action: fae.activity_log_filter_path do
     == fae_filter_select :user, options: Fae::User.all.map { |u| [u.full_name, u.id] }
-    == fae_filter_select :model, options: @items.unique_changeable_types
+    == fae_filter_select :model, options: Fae::Change.unique_changeable_types
```
jamesmk: @jasonfine What is the issue this is solving? The idea is the filers should only display available `changeable_types`, so running this on the collection should work.

jasonfine: since @items is a paginated set there's only X number of them, so we're only plucking what's available from the @items set and not from the whole pool, which the class/table would contain

PR #281, opened 2017-07-20
Incomplete trumbowyg library

jeremycrosbie:
Using `html: true` in a form doesn't work because the WYSIWYG plugin isn't specified in application.js.  Also, the trumbowyg library isn't fully included--it's missing the icon images for the various controls.


jeremycrosbie:
Saw that this is supposed to be loaded via a different method than I presumed.

PR #282, opened 2017-07-22
Ruby 2.4 compatibility

Spone:
I added `::` before `File` in generators to fix the collision with `Fae::File`


jamesmk:
Thanks @Spone!

PR #285, opened 2017-08-01
authorize global search results

jasonfine:
ended up having to put the auth check on the `records` set instead of trying to do it on the `all_models` collection because it won't work with content blocks on that level.

### DIFF
--- a/app/controllers/fae/utilities_controller.rb
+++ b/app/controllers/fae/utilities_controller.rb
@@ -54,7 +54,25 @@ module Fae
       all_models.each do |m|
         records += m.fae_search(query)
       end
-      records
+      authorize_records(records)
+    end
+
+    def authorize_records(records)
+      users_role = current_user.role.name.downcase
+      authorized_records = []
+
+      records.each do |record|
+        tableized_model = record.class.name.tableize
+        role_group_for_model = Fae::Authorization.access_map[tableized_model]
+
+        if role_group_for_model.present?
+          authorized_records << record if role_group_for_model.include?(users_role)
+        else
+          authorized_records << record
+        end
+      end
+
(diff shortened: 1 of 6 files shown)


app/controllers/fae/utilities_controller.rb
```diff
@@ -54,7 +54,25 @@ def records_by_display_name(query)
       all_models.each do |m|
         records += m.fae_search(query)
       end
-      records
+      authorize_records(records)
+    end
+
+    def authorize_records(records)
```
jamesmk: @jasonfine I have some performance concerns with this method since we're looking up the authorization from the model per record. Since we're already looping through models via `all_models`, what if we filtered that down to `authorized_models`? I think that would drastically cut down on the `Fae::Authorization.access_map` look ups. What do you think?

spec/features/global_search_spec.rb
```diff
@@ -34,4 +34,34 @@
     end
   end
 
+  # see dummy app's authorization concern for auth mapping used here
+  scenario "search results get authorized", js: true do
```
jamesmk: 👍 🎉

app/controllers/fae/utilities_controller.rb
jasonfine: that's originally the route i went with this, then realized it doesn't work for situations where  users are limited from seeing arbitrary fae::pages, ill take another look in the AM and see what might be possible here

spec/features/global_search_spec.rb
jasonfine: ok moved all this to work on all_models, i didn't realize it was already working on pages because the static items in the global search are built from the already-authorized nav structure. win-win

PR #286, opened 2017-08-03
initial pass for asset cloning

jasonfine:
1) rip apart asset_cloner.rb to your heart's content - it works well as-is though.

2) the existing cloning implementation did not work for has_one's, and there were no tests for has_one's cloning. during this i realized has_one's are basically never used in any of our apps outside of file attachments, so i can see why that was overlooked.

3) had a *really* difficult time getting the fae file factories to work in the context of a capybara feature spec (where the rest of the cloning tests are).  after hours of wheel-spinning i threw in the towel and moved the asset cloning spec into a request spec and it works without issue there.


jasonfine:
dealing with failing test

PR #295, opened 2017-08-07
Refs #60171 - Helpful multiselect actions

aaronransley:
Internal issue: #60171

Adds "Select All" and "Deselect All" options to multiselect dropdowns

@jamesmk I'm not sure why the commit log for this one is so large, but the actual file modifications are very narrow. I'm happy to do some additional work with rebases or perhaps cherry picking my changes onto a fresh branch if that makes things cleaner ✨


jamesmk:
@aaronransley I think you must've rebased some commits on your branch that were on master as well, assigning them new SHAs. These commits will be duplicated in master's history, so would you kindly cherry-pick your commits onto a clean branch?

thank you ❤️

PR #299, opened 2017-08-10
Refs #60171 - Adds a helpful actions to multiselects

aaronransley:
**Internal issue:** https://issues.afinedevelopment.com/issues/60171

* Adds "SELECT ALL" as an option to multiselect fields.
* When a multiselect field has an item is selected, "Deselect All" now appears below the field.

Can be verified [on stage](http://stage.fae.afinedevelopment.com/admin/releases/3/edit#associations) — check the Acclaims multiselect.

### DIFF
--- a/app/assets/javascripts/fae/form/_form.js
+++ b/app/assets/javascripts/fae/form/_form.js
@@ -7,6 +7,9 @@
  */
 Fae.form = {
   ready: function() {
+    // Mutate DOM to support two column labels for all standard inputs
+    this.makeTwoColumnLabels();
+
     this.dates.init();
     this.text.init();
     this.select.init();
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/fae/form/_form.js
```diff
@@ -23,19 +26,17 @@ Fae.form = {
...
   },
 
   makeTwoColumnLabels: function() {
     $('.input label').each(function() {
       var $element = $(this);
+      var has_no_helper_text = false;
 
       // Bail if we cannot find any helper_text
-      if (!$element.find('.helper_text').length) { return; }
+      if (!$element.find('.helper_text').length) { has_no_helper_text = true; }
```
jamesmk: @aaronransley I feel like this implementation detaches the logic a bit and extends an already long function. What do you think about calling another function here?

```
if (!$element.find('.helper_text').length) { addNoHelperClass($element); }
```

PR #305, opened 2017-08-14
Add TravisCI testing

cpitkin:
Testing is now done in TravisCI.


jamesmk:
@cpitkin I think you want to merge `test-fixes` into `master`. :) Doesn't look like I can switch the direction via edit so I'll reopen a new one.

thanks

PR #307, opened 2017-08-15
Initial tutorial integrating Devise and Fae

arinhouck:
This will assist you in using Fae with exiting Devise controllers and views.


jamesmk:
Awesome write up @arinhouck, thank you!

PR #317, opened 2017-08-23
refs#69450-don't submit the form if errors are present

JuliannaGreen:
Currently the form is submitting during validation. Validation should stop the form from submitting and only submit if no errors are present.
This is a temporary hot fix for this problem. 

This file should be refactored at large sooner rather than later.

### DIFF
--- a/app/assets/javascripts/fae/form/_validator.js
+++ b/app/assets/javascripts/fae/form/_validator.js
@@ -26,7 +26,7 @@ Fae.form.validator = {
   formValidate: function () {
     var _this = this;
     FCH.$document.on('submit', 'form', function (e) {
-      _this.is_valid = true;
+      e.preventDefault();
 
       // Scope the data-validation only to the form submitted
       $('[data-validate]', $(this)).each(function () {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/fae/form/_validator.js
```diff
@@ -51,6 +51,13 @@ Fae.form.validator = {
       if ($(".field_with_errors").length) {
         $('.alert').slideDown('fast').removeClass('hide').delay(3000).slideUp('fast');
       }
+
+      if ($('.invalid').length) {
```
jamesmk: @JuliannaGreen I have a couple concerns with this change:
I don't think we should be sniffing something so general as an `invalid` class. There are cases where we have multiple forms on the page (as with nested forms) which may be prevented from submitting if the parent form has errors. If this class is reliable at determining the validity of the form, then maybe scoping the lookup to the form would be better?

If we can't count on the `is_valid` flag, likely that's where the bug is. I believe it's only purpose is to determine whether or not to submit the form, so you're method here would likely mean stripping out the flag altogether. However, I would bet the issue is in properly setting the flag during the `_judgeIt()` call.

jamesmk: Also, the Travis build is failing, you can click to view the details to see which tests are failing. At first glance it looks like it's around nested forms.

PR #328, opened 2017-09-14
69514 decouple pagination from filters

andrewdang17:


app/assets/javascripts/fae/form/_filtering.js
```diff
@@ -37,8 +37,16 @@ Fae.form.filtering = {
...
     // hardcode _this because this === Fryr object
     var _this = Fae.form.filtering;
-    var post_url = _this.$filter_form.attr('action');
-    var $results_table = ($(".js-results-table").length) ? $(".js-results-table").first() : _this.$filter_form.next('table');
+
+    if (_this.$filter_form.length) {
+      post_url = _this.$filter_form.attr('action');
+      $current_table = _this.$filter_form.next('table');
+    } else {
+      post_url = _this.$pagination.data('filter-path');
```
jamesmk: The data attribute below is snake cased and this uses a hyphen. Is this working? I would assume it wouldn't.

docs/features/filtering.md
```diff
@@ -175,14 +175,13 @@ resources :people do
...
-
-Then call the `fae_paginate` helper under the table:
+Add a `paginated` class to the items table, then call the `fae_paginate` helper under the table:
 
 `app/views/admin/people/index.html.slim`
 ```slim
 / app/views/admin/people/index.html.slim
 
-table
+table.paginated
```
jamesmk: I would like things to work automatically with as little integration steps as possible. Instead of using this class to look up the table, what do you think about not requiring it and looking up the sibling table above `fae_paginate`?

app/assets/javascripts/fae/form/_filtering.js
andrewdang17: Yeah this works. Rails converts it to kebab case since symbols can't have hyphens. If I do `data: { filter-path: '' }` I get a syntax error

docs/features/filtering.md
andrewdang17: Sounds good to me!

PR #333, opened 2017-10-12
style flash message errors appropriately

tshedor:


jamesmk:
@tshedor I'm curious where you saw this issue?

tshedor:
@jamesmk Ref integrity. I threw a 

```ruby
has_many :types, dependent: :restrict_with_error
```

on my model and the flash message when I tried to delete something was unstyled

PR #335, opened 2017-10-16
refs #69450 - wait for all validation checks to complete before submitting form

jamesmk:
Unique validation checks on forms require Judge to do an AJAX call. However in some cases the form is checked for validity and submitted before the AJAX response triggers the callback.

Since the AJAX call is from a 3rd party it's difficult to use callbacks or monitor the XHR object. This solution tracks the number of validations checks every 50ms until they all have responded. Then the form will submit or display errors accordingly.

### DIFF
--- a/app/assets/javascripts/fae/form/_validator.js
+++ b/app/assets/javascripts/fae/form/_validator.js
@@ -44,7 +44,7 @@ Fae.form.validator = {
         _this.validation_test_count = 0;
 
         // Scope the data-validation only to the form submitted
-        $('[data-validate]', $(this)).each(function () {
+        $('[data-validate]', $this).each(function () {
           if ($(this).data('validate').length) {
             _this.validations_called++;
             _this._judgeIt($(this));
@@ -52,7 +52,7 @@ Fae.form.validator = {
         });
 
         // Catch visible errors for image/file inputs hitting the fae config file size limiter
-        $('.input.file', $(this)).each(function () {
+        $('.input.file', $this).each(function () {
           if ($(this).hasClass('field_with_errors')) {
             _this.is_valid = false;
           }
--- a/app/assets/javascripts/fae/form/_validator.js
+++ b/app/assets/javascripts/fae/form/_validator.js
@@ -8,6 +8,9 @@
 Fae.form.validator = {
 
   is_valid: '',
+  validations_called: 0,
+  validations_returned: 0,
+  validation_test_count: 0,
 
   init: function () {
     if (FCH.exists('form')) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/fae/form/_validator.js
```diff
@@ -21,37 +24,83 @@ Fae.form.validator = {
...
+        e.preventDefault();
+
+        // set defaults
+        _this.is_valid = true;
+        _this.validations_called = 0;
+        _this.validations_returned = 0;
+        _this.validation_test_count = 0;
+
+        // Scope the data-validation only to the form submitted
+        $('[data-validate]', $(this)).each(function () {
```
tshedor: `$(this)` here is actually `$this`

app/assets/javascripts/fae/form/_validator.js
```diff
@@ -21,37 +24,83 @@ Fae.form.validator = {
...
+          if ($(this).data('validate').length) {
+            _this.validations_called++;
+            _this._judgeIt($(this));
+          }
+        });
+
+        // Catch visible errors for image/file inputs hitting the fae config file size limiter
+        $('.input.file', $(this)).each(function () {
```
tshedor: This `$(this)` is also `$this`

app/assets/javascripts/fae/form/_validator.js
```diff
@@ -21,37 +24,83 @@ Fae.form.validator = {
...
-      });
 
-      if (_this.is_valid === false) {
-        Fae.navigation.language.checkForHiddenErrors();
-        FCH.smoothScroll($('#js-main-header'), 500, 100, 0);
-        e.preventDefault();
+      } else {
+        // check again if it hasn't run more than 50 times
+        // (to prevent against infinite loop)
+        if (_this.validation_test_count < 50) {
```
tshedor: This feels uncomfortably hacky. I'm assuming the `50` was arbitrarily selected, but setting a timeout of `2500` would also be an arbitrary number. However, since it is based on AJAX that may or may not send, it's the best solution. 

Ultimately, the ideal solution would be to receive a Promise (or equivalent, you're more familiar with Judge's internals than I am) from Judge that resolves after all validations have returned from the server.

jamesmk: Yep 50 is an arbitrary number, but it is merely a fallback to prevent an infinite loop.

I looked into a callback or promise on Judge's side, but I feel it would take a lot of supporting code to pass it to Fae. The AJAX can is written in vanilla JS and the call itself, where I'd usually receive the promise is in Judge as well.

jamesmk:
I'll address @tshedor's comments and the tests I broke 😵

Issue #336, opened 2017-10-18
Consolidating form_header in v2.0


tshedor:
@jamesmk Master of deletions here, proposing a deletion.

The current generator spits this out:

```slim
header.content-header.js-content-header
  == render 'fae/shared/form_header', header: @klass_name
== render 'fae/shared/errors'
```

This is in _a lot_ of files. Since this is a breaking version, I'd like to propose expanding the `form_header` partial to combine everything.

```slim
== render 'fae/shared/form_header', header: @klass_name
```

On content blocks, it'd look like this

```slim
== render 'fae/shared/form_header', header: @item, f: f, errors: false
```

**Cons**
Upgrading from v1.x to v2.0 will require manual remove of all `header.content-header.js-content-header`. 

Anyone using a customized `form_header` will have to add the wrapper `header.content-header.js-content-header` to their customized partial. Looking at you, J.

**Pros**
Should a redesign occur in the future, HTML in the form header will not be a concern because it is managed in the partial. The more in the partials the better. With partials, there's less concern for breaking changes and less need to maintain older HTML when a JS file is updated, when a CSS class is changed, when extra functionality needs to be added to the header and the HTML is in the way.

Two less lines per `_form.html.slim`.

jamesmk:
I'm a fan of this 👍

PR #342, opened 2017-11-01
Consolidate form header

tshedor:
Satisfying the proposal from  #336

### DIFF
--- a/app/views/fae/shared/_form_header.html.slim
+++ b/app/views/fae/shared/_form_header.html.slim
@@ -10,30 +10,35 @@ ruby:
   title              ||= "#{params[:action]} #{header_as_string}".titleize
   breadcrumb           = !local_assigns[:breadcrumbs].eql?(false)
   languages          ||= false
+  hide_flash         ||= false
 
-- if breadcrumb
-  == render 'breadcrumb', title: title
-- else
-  = form_header header
+header.content-header.js-content-header
+  - if breadcrumb
+    == render 'breadcrumb', title: title
+  - else
+    = form_header header
 
-.content-header-buttons
-  - if languages
-    - selected = current_user.language || ''
+  .content-header-buttons
+    - if languages
+      - selected = current_user.language || ''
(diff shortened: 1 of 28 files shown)


app/views/fae/shared/_form_header.html.slim
```diff
@@ -10,30 +10,35 @@ ruby:
   title              ||= "#{params[:action]} #{header_as_string}".titleize
   breadcrumb           = !local_assigns[:breadcrumbs].eql?(false)
   languages          ||= false
+  hide_flash         ||= false
```
jamesmk: This reads as a double negative, I would prefer `show_flash` with the default being true. Of course you can't pass in `false` with the standard or equauls, I think you have to do something like:

```ruby
show_flash = true if show_flash.nil?
```

lib/generators/fae/templates/views/static_page_form.html.slim
```diff
@@ -1,6 +1,5 @@
 = simple_form_for @item, url: fae.update_content_block_path(slug: @item.slug), method: :put do |f|
-  header.content-header.js-content-header
-    == render 'fae/shared/form_header', header: @item
+  == render 'fae/shared/form_header', header: @item, hide_flash: true
```
jamesmk: I know we'll likely need `hide_flash` (or `show_flash`) to support legacy  apps, but I'm wondering if we should have available on all forms by default. Can you think of why it would be omitted on the static page forms?

tshedor: Did some digging, and the errors partial was added to the `_form.html.slim` generator in 057485a5 . All the models that didn't have it (e.g. `acclaims` or `coaches`) were generated before 057485a5 . So I'll go back and set those to show flash messages in this PR.

However, for the static_pages, those have never had the flash messages partial. Those errors are handled inline, per form field. I guess it would only make it more cohesive if they all had flash messages.

PR #348, opened 2017-11-09
add index_table partial

tshedor:
A partial for indexes will remove boilerplate code and streamline the updating process when markup is affected.

I haven't written tests because I wanted to get feedback on the concept and to confirm the local variables passed through to it.

### DIFF
--- a/app/helpers/fae/nested_form_helper.rb
+++ b/app/helpers/fae/nested_form_helper.rb
@@ -3,15 +3,19 @@ module Fae
 
     def th_columns(attribute)
       if (attribute.is_a?(Hash) && attribute[:title])
+        attribute_sort = attribute[:sort_by]
         attribute = attribute[:title]
         attribute_title = attribute
       else
         attribute = :modified if ['updated_at', 'modified_at'].include? attribute.to_s
         attribute_title = attribute.to_s.titleize
+        attribute_sort = nil
       end
 
       th_class = '-action-wide' if [:on_production, :on_prod, :on_stage, :updated_at, :created_at, :modified].include?(attribute)
-      content_tag(:th, class: th_class) do attribute_title end
+      th_sort = attribute_sort
+      th_sort = attribute.to_s if [:updated_at, :created_at, :modified].include?(attribute)
+      content_tag(:th, class: th_class, data: { sorter: th_sort }) do attribute_title end
     end
 
     def td_columns(params)
--- a/app/views/fae/shared/_index_table.slim
+++ b/app/views/fae/shared/_index_table.slim
@@ -1,0 +1,74 @@
+ruby:
+  require_locals ['edit_path'], local_assigns
+  items         ||= @items || []
+  cols          ||= []
+  sort_columns  ||= true
+  sort_rows       = local_assigns[:sort_rows].eql?(true)
+  table_class   ||= ''
+  image         ||= nil
+  edit_label    ||= 'Name'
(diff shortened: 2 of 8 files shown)


jamesmk:
I have some reservations about his change. I get that it DRYs up some of the index views and gives us more ability to update markup behind the scenes, but I can't help but feel like it takes Fae further away from feeling Rails-like. This is such a high-touch view, I'd like any Rails dev to be able to jump in and have a sense of what's going on. This change further obfuscates that behind Fae's growing DSL.

Also we've done a lot of custom work in some of our index views, which will likely require an overwrite with this change. I don't know if that option would be obvious to other Fae users.

I'm curious what others have to say on this. And thanks as always @tshedor!

tshedor:
I would disagree that it makes it less Rails-like. The ease and appeal of Rails is all the code it tucks away. Admittedly, it's too much magic at times, but I believe this partial strikes the proper balance.

As far as obfuscation, this pattern is similar to existing [nested tables](https://github.com/wearefine/fae/blob/086f1cd8fe4e7f4ee3918c192f56e4f88adf2dc7/spec/dummy/app/views/admin/cats/index.html.slim) and [index forms](https://github.com/wearefine/fae/blob/086f1cd8fe4e7f4ee3918c192f56e4f88adf2dc7/spec/dummy/app/views/admin/cats/index.html.slim). I see this PR being closer to Fae's existing pattern than deviating from it. 

There are some indexes, such as the [releases form](https://github.com/wearefine/fae/blob/d024e79061365b335e8737825101763a6e2b6938/spec/dummy/app/views/admin/releases/index.html.slim), that require more than this partial can provide, and in those cases I see no issue with using traditional markup ([as noted](https://github.com/wearefine/fae/pull/348/files#diff-cbb99751db1eb0d01341844c54107437R67)). But I think for the vast majority of scaffolded admins, this partial is efficient _and_ sufficient.

However, I'm also curious as to what others have to say and was hoping for further discussion on this PR by now. I don't mind if it perpetually remains open.

andrewdang17:
Gonna have to go with @jamesmk on this, I think this is too much abstraction that I'd end up referring to the docs more because of it. I'd also prefer more control over the markup of the table and think the boilerplate is worth it in this case.

jamesmk:
@tshedor I get that the abstraction is close to the existing nested tables, but I have a secret: I hate those things. When we need to support custom cols, etc, the partials go overly complex and the DSL became even more obscure. I'd love to find a cleaner solution for those.

JuliannaGreen:
I honestly find repeating markup in index views to be tedious. I find an index partial to be somewhat appealing for that reason. I do worry about it introducing complications for customization though. 

Another point is that this locks the user into using this layout or system for their index views, but like Tim points out we are already forcing this with nested index views anyway.

johnnyshields:
I might change the name to something like `simple_table` to indicate that it's meant to cover 80% of the use cases, but complex cases will still require custom code. In the case of my app I would be able to use this for about 80% of my pages so I would consider a benefit to having this, especially for upgrading.

jamesmk:
After further internal discussion I have decided to reject this. I find the nested table partial inflexible, we are often wrestling with it on our internal projects. This shows in features we've added to help "customize" it, like custom attrs and titles (https://github.com/wearefine/fae/blob/master/docs/helpers/partials.md#custom-titles-in-cols).

I don't want the index table to take the same route.

PR #350, opened 2017-11-11
Loosen restriction on rmagick gem

johnnyshields:


jamesmk:
@johnnyshields I would rather keep it locked to a minor version and do a very intentful upgrade. Was there a bug or feature specifically related to this change? FYI we'll likely be moving to `mini_magick` in v2.0 by EOY.

thanks

johnnyshields:
Moved to mini_magick so not needed

PR #352, opened 2017-11-12
Fix Rails migrations

johnnyshields:
Rails 5.1 requires AR::Migration version number in the migrations, otherwise they will fail.


jamesmk:
@johnnyshields Thanks for the PR, however this change isn't compatible with Rails 4 which Fae v1 still supports. We're are working on Fae v2.0 now with Rails 5.1 support (dropping Rails 4 support). If you'd like to reopen a PR into the v2.0 branch that's more compatible, I'd be happy to merge it.

thanks

PR #354, opened 2017-11-13
limit image alt and caption fields to 125 chars

jasonfine:


jamesmk:
@jasonfine I can see why we'd set a character limit on the `alt` as it seems to be a best practice accessibility wise (http://accessibility.psu.edu/images/imageshtml/). I don't think we should limit `caption` by default as that can be used as an aux text field associated with the image. It would be best left flexible for different use cases.

Another slight concern I have about this is that if someone upgrades to this change, current content might not fit within these restrictions, making once valid forms invalid. It could almost be considered non-backwards compatible.

jasonfine:
just going to override in app for now

jamesmk:
@jasonfine What do you think about opening a PR for v2.0 with character limit for `alt`?

PR #360, opened 2017-12-04
allow blank labels for fae_content_form

andrewdang17:
This is just a quick fix so I can get ticket done for a client this week. I think later on all that logic in `_content_form.html.slim` should be refactored out to be separate from the partial so that you just have to pass  an `options` in rather than every possible variable since everything is a bit dependent on each other right now and relies on a lot of conditionals.

### DIFF
--- a/app/views/fae/application/_content_form.html.slim
+++ b/app/views/fae/application/_content_form.html.slim
@@ -1,6 +1,6 @@
 ruby:
   require_locals     ['attribute', 'f'], local_assigns
-  blank_label          = !label || label == '' unless label.nil?
+  has_label            = ![false, ''].include?(local_assigns[:label])
   label              ||= attribute.to_s.titleize
   required           ||= f.object.send(attribute).is_required?(f.object.class)
   helper_text        ||= attempt_common_helper_text(attribute)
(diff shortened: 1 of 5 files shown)


app/views/fae/application/_content_form.html.slim
```diff
@@ -1,5 +1,6 @@
 ruby:
   require_locals     ['attribute', 'f'], local_assigns
+  blank_label          = !label || label == '' unless label.nil?
   label              ||= attribute.to_s.titleize
```
jamesmk: @andrewdang17 We're starting to spread the label logic all over this file. The `label_adjusted` assignment on line 13 is where the full label get's built, so I think that's where the conditional logic should live. There should be no need to execute that logic if we're just setting it to false later.

app/views/fae/application/_content_form.html.slim
```diff
@@ -1,5 +1,6 @@
 ruby:
   require_locals     ['attribute', 'f'], local_assigns
+  blank_label          = !label || label == '' unless label.nil?
```
jamesmk: I find this line difficult to interpret. If we focus on supporting `label: false`, then it could look something like:

```ruby
has_label = !local_assigns[:label].eql?(false)
```

This is a pattern we've using in other partials too.

app/views/fae/application/_content_form.html.slim
andrewdang17: I agree I just did it this way because I didn't want to add more `if/unless` statements to lines 12-22 or have to wrap the whole thing in an if statement

app/views/fae/application/_content_form.html.slim
andrewdang17: Nice that looks better. `label: false` was the only way to remove the label for simpleform hence the check for false or an empty string

andrewdang17: @jamesmk what do you about using `![false, ''].include?(local_assigns[:label])` instead?

The problem with your way is that if `label: false`, then `!local_assigns[:label]` will return true when really it should be false.

andrewdang17:
@jamesmk updated

PR #361, opened 2017-12-06
Added support for decorator pattern

olhor:


jamesmk:
Thanks for the PR @olhor!

PR #366, opened 2017-12-11
move some fae_content_form logic out of partial

andrewdang17:
There was a bit too much conditional and key/value checking and I thought I'd move some of the `fae_content_form` logic out of the partial and into a class since logic shouldn't be in a view. Let me know what you think.


jamesmk:
This has potential for a future refactoring and I like the concept. But since this would be the only partial using this abstraction I feel like it adds more technical debt as is.

PR #367, opened 2017-12-12
Prevent breakage of the asset pipeline

olhor:
I want to change the way how user defined styles can be loaded from `app/assets/stylesheets/fae.scss` by providing the ability to support usage of sprockets directives within `app/assets/stylesheets/fae.scss`. The scss `@import` directive does not allow to do that.

### DIFF
--- a/app/assets/stylesheets/fae/application.css
+++ b/app/assets/stylesheets/fae/application.css
@@ -5,4 +5,5 @@
  *= require fae/vendor/simplemde.min
  *= require fae/vendor/trumbowyg
  *= require fae/base
+ *= require fae
  */
--- a/app/assets/stylesheets/fae/base.scss
+++ b/app/assets/stylesheets/fae/base.scss
@@ -6,9 +6,6 @@
   'globals/imports/extends'
 ;
 
-// User defined stuff
-@import 'fae';
-
 // Fonts, vendor, and root HTML tag styles
 @import
   'globals/fonts',


tshedor:
@olhor Without `@import 'fae'`, variables are not passed through to the rest of the Fae styles. For instance, `$c-custom-highlight` provides the dominant color throughout the interface and requiring the file with Sprockets would ignore that override. 

Other overrides that would not be added to the installation include custom grays, fonts, and flash background colors.

olhor:
So maybe just move the line `$c-custom-highlight` from `fae.scss` to `app/assets/stylesheets/fae/base.scss` and add sprockets directives as I did. This will leave `fae.scss` sprockets compatible.
I understand that it will break backwards compatibility but proper upgrade notes can be written.

The other approach would be to require other custom user file with sprockets support and leave `fae.scss` as it is.

tshedor:
Moving child app variables to `base.scss` would require the file to be filled with boilerplate `@import`s. Those imports are likely to change and cause headaches during future upgrades.

I like the idea of a second file that would support Sprockets, although ultimately I think that would be confusing to have *two* Fae files in the assets folder of the child app when the `.scss` file already covers nearly all use cases.

What do you think about simply overriding the layout file in the child app? This file could change in the future, but if you're adding external libraries using Sprockets, you're already playing with a customized setup.

```slim
/ app/views/layouts/fae/application.html.slim
doctype html
html
  head
    title = page_title
    ...
    == stylesheet_link_tag 'fae/application', media: 'all'
    == stylesheet_link_tag 'fae-sprockets', 
    == javascript_include_tag 'fae/application'
    ...
```

olhor:
I wanted to avoid overriding the view but I suppose this is only thing that can be done right now.

PR #374, opened 2017-12-21
fae_display_field is supposed to be scoped to the instance not the class

johnnyshields:
Please review/comment

I think maybe this needs to autodetect `name` and `title` (maybe also `label`)


jamesmk:
You're right, thanks for the PR!

PR #404, opened 2018-04-25
Bump judge gem version

johnnyshields:
Old Judge was causing issues for me, specifically we need this fix:
https://github.com/joecorcoran/judge/commit/1265b82288bc072ff01c41d2bb17291ae9866d8e


jamesmk:
@johnnyshields Thanks for the PR, but I'm going to close this as it duplicates #401. I'm doing some final production checks on v2 this week and then will officially release it. If you'd like to use it immediately, you can link `fae-rails` to the `v2.0` branch.

PR #412, opened 2018-06-26
Clarifying supported Rails versions in README

kurtome:
It appears you already support Rails 5.2 in the latest version of Fae, at least that's what is referenced int he Upgrading documentation.


jamesmk:
Good catch @kurtome, thanks!

PR #413, opened 2018-06-28
left colour picker offset when sidenav present

friendofdog:
When sidenav is present, the colour picker breaks. Picker's ```left``` CSS value is a static ```30px```, which puts it out of position relative to the form. Changed this to dynamic value.


jamesmk:
Thanks for the bugfix @friendofdog!

PR #417, opened 2018-07-08
Minor docs fix

MrSkinny:
Syntax for Markdown links was incorrectly inverted and causing breakage on https://www.faecms.com/documentation/installation-index


jamesmk:
Thanks @MrSkinny!

PR #419, opened 2018-07-18
Nested AJAX form validation failure fix

jclemans:
## Summary ##
Adds a call to re-bind form field validation events after a nested AJAX form is loaded.

## Background ##
The way that _validator.js is set up runs through a loaded page and binds all the form fields that have `.data-validate` on them. But a nested form that is loaded via AJAX does not get addressed, so this change re-runs the binding action on the form after the new form fields are loaded.

Related issues:

- Redmine 75403
- Flow 35293743

### DIFF
--- a/app/assets/javascripts/fae/form/_ajax.js
+++ b/app/assets/javascripts/fae/form/_ajax.js
@@ -67,7 +67,7 @@ Fae.form.ajax = {
       }
 
       // Bind validation to nested form fields added by AJAX
-      Fae.form.validator.bindValidationEvents($scope = $('.nested-form'));
+      Fae.form.validator.bindValidationEvents($('.nested-form'));
 
       // Reinitialize form elements
       Fae.form.dates.initDatepicker();
--- a/app/assets/javascripts/fae/form/_ajax.js
+++ b/app/assets/javascripts/fae/form/_ajax.js
@@ -66,8 +66,8 @@ Fae.form.ajax = {
         $wrapper.find('.input.file').fileinputer();
       }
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/fae/form/_ajax.js
```diff
@@ -66,6 +66,9 @@ Fae.form.ajax = {
         $wrapper.find('.input.file').fileinputer();
       }
 
+      // Re-bind form validations to catch new AJAX loaded fields
+      Fae.form.validator.bindValidationEvents();
```
jamesmk: @jclemans We chatted IRT about adding the ability to optionally scope this method so we don't have to rebind all form event handlers every time nested form AJAX is called. Adding as a review item, let me know if you hit any walls on it.

thanks

app/models/fae/static_page.rb
```diff
@@ -42,15 +42,15 @@ def self.setup_dynamic_singleton
           languages.each do |lang|
             # Save with suffix for form fields
             define_association("#{name}_#{lang}", type)
-            define_validations("#{name}_#{lang}", type, value[:validates]) if value.try(:[], :validates).present?
```
jclemans: @jamesmk - not sure why these changes static_page diffs are showing up. I just pulled from master to bring my branch up to date with your recent changes.

jclemans:
@jamesmk - updated to fix the failing capybara tests / IE11 issue. Just the one flickering date_sort_spec failing now.

jamesmk:
thanks @jclemans!

PR #421, opened 2018-08-17
Nested ajax form fixes for file attachment fields

jclemans:
Nested form validations are not triggering in cases where a form field is not touched by a blur event.  A user can fill click submit on an empty or partially empty set of nested forms and no form validations will be triggered. This fixes by passing the nested-form as a scope so that a submit click listener can be initialized on the nested form.

I also made some changes to fix a minor bug with the delete link on the file input in cases where the object was not saved successfully. If there are form errors, there is no `file.id` so the delete file path would break the form during rendering. I added an alternate simple <a> tag link with some JS to handle clearing the file input in this circumstance.

This work is related to Redmine issue 75403

### DIFF
--- a/app/assets/javascripts/fae/form/_ajax.js
+++ b/app/assets/javascripts/fae/form/_ajax.js
@@ -17,7 +17,6 @@ Fae.form.ajax = {
     this.addCancelLinks();
 
     this.imageDeleteLinks();
-    this.fileDeleteLinks();
     this.htmlListeners();
 
     this.deleteNoForm();
@@ -289,25 +288,22 @@ Fae.form.ajax = {
     });
   },
 
-  fileDeleteLinks: function() {
-    $('body' ).on('click', 'a.js-file-clear', (function(e) {
-
-      e.preventDefault();
-      var $this = $(this);
-
-      if (confirm('Are you sure you want to clear this file?')) {
-        $this.parent().next().show();
-        $this.parent().hide();
-      }
-    }));
-  },
-
   /**
    * Attaching click handlers to #js-main-content to allow ajax replacement
    * @todo Clean this up, moving listeners into their respective component classes (select, checkbox, etc.)
    */
   htmlListeners: function() {
     $('#js-main-content, .login-form > form')
+
+      /**
+       * For the delete button on file input
+       */
+      .on('click', '.js-file-clear', function(e) {
+        e.preventDefault();
+        var $parent = $(this).parent();
+        $parent.next().show();
+        $parent.hide();
+      })
 
       /**
        * For the yes/no slider
--- a/app/assets/javascripts/fae/form/_ajax.js
+++ b/app/assets/javascripts/fae/form/_ajax.js
@@ -17,6 +17,7 @@ Fae.form.ajax = {
     this.addCancelLinks();
 
     this.imageDeleteLinks();
+    this.fileDeleteLinks();
     this.htmlListeners();
 
     this.deleteNoForm();
@@ -286,6 +287,19 @@ Fae.form.ajax = {
         $this.parent().hide();
       }
     });
+  },
+
+  fileDeleteLinks: function() {
+    $('body' ).on('click', 'a.js-file-clear', (function(e) {
+
+      e.preventDefault();
+      var $this = $(this);
+
+      if (confirm('Are you sure you want to clear this file?')) {
+        $this.parent().next().show();
+        $this.parent().hide();
+      }
+    }));
   },
 
   /**
--- a/app/assets/javascripts/fae/form/_ajax.js
+++ b/app/assets/javascripts/fae/form/_ajax.js
@@ -10,6 +10,7 @@ Fae.form.ajax = {
   init: function() {
     this.$addedit_form = $('.js-addedit-form, .js-index-addedit-form');
     this.$filter_form = $('.js-filter-form');
+    this.$nested_form = $('.nested-form');
 
     this.addEditLinks();
     this.addEditSubmission();
@@ -67,7 +68,7 @@ Fae.form.ajax = {
       }
 
       // Bind validation to nested form fields added by AJAX
-      Fae.form.validator.bindValidationEvents($('.nested-form'));
+      Fae.form.validator.bindValidationEvents(this.$nested_form);
 
       // Reinitialize form elements
       Fae.form.dates.initDatepicker();
@@ -79,7 +80,7 @@ Fae.form.ajax = {
       Fae.form.checkbox.setCheckboxAsActive();
 
       // validate nested form fields on submit
-      Fae.form.validator.formValidate($('.nested-form'));
+      Fae.form.validator.formValidate(this.$nested_form);
 
       $wrapper.find('.hint').hinter();
     });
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/fae/form/_validator.js
```diff
@@ -84,9 +88,11 @@ Fae.form.validator = {
 
           $this.submit();
         } else {
-          // otherwise scroll to the top to display alerts
+          // otherwise scroll to the top to display alerts (unless in a nested form scope)
           Fae.navigation.language.checkForHiddenErrors();
-          FCH.smoothScroll($('#js-main-header'), 500, 100, 0);
+          if (typeof($scope) === 'undefined') {
```
jclemans: This prevents the page from scrolling all the way back to the top of the parent form when you clear out a selected item in a nested form at the bottom of the page.

app/assets/javascripts/fae/form/_ajax.js
```diff
@@ -129,7 +133,7 @@ Fae.form.ajax = {
           } else if ($html.hasClass('nested-form')) {
 
             // we're returning the form due to an error, just replace the form
-            $this.find( '.nested-form' ).replaceWith(data);
```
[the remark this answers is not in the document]
tshedor: A bit of history: this nested-form rendering has been passed down from many generations. I believe this is the fifth iteration and you are its fourth author. In short, [your code is always awesome and the other people's code is not awesome](https://youtu.be/PLHln7wHgPE?t=5m40s). 

While the way this is currently written appears inane, it is likely done for a specific instance (like AJAX'd inputs on index pages) which inevitably became the Gospel of Hack for all use cases. 

If you're feeling bold, I would suggest revisiting the whole function and breaking it down even further (while still paying attention to the function's legion, chimeric use cases).

app/assets/javascripts/fae/form/_ajax.js
```diff
@@ -285,6 +289,19 @@ Fae.form.ajax = {
     });
   },
 
+  fileDeleteLinks: function() {
+    $('body' ).on('click', 'a.js-file-clear', (function(e) {
+
+      e.preventDefault();
+      var $this = $(this);
```
tshedor: Instead of caching `$this`, just cache `$parent = $(this).parent()`.

app/assets/javascripts/fae/form/_ajax.js
```diff
@@ -285,6 +289,19 @@ Fae.form.ajax = {
     });
   },
 
+  fileDeleteLinks: function() {
+    $('body' ).on('click', 'a.js-file-clear', (function(e) {
```
tshedor: Use `FCH.$body`

tshedor: Is there any reason for wrapping the third argument (the anonymous function) in an uninvoked manner like this? Legitimate question, I've only seen `(function` when it's an IIFE.

app/assets/javascripts/fae/form/_validator.js
```diff
@@ -26,10 +26,13 @@ Fae.form.validator = {
   /**
    * Validate the entire form on submit and stop it if the form is invalid
    */
-  formValidate: function () {
+  formValidate: function ($scope) {
     var _this = this;
 
-    FCH.$document.on('submit', 'form:not([data-remote=true])', function (e) {
+    if (typeof($scope) === 'undefined'){
```
tshedor: FWIW Frob Core has a [`setDefault`](https://github.com/wearefine/fae/blob/master/app/assets/javascripts/fae/vendor/frob_core_helpers.js#L317-L319) for better legibility

app/assets/javascripts/fae/form/_validator.js
```diff
@@ -69,7 +72,7 @@ Fae.form.validator = {
    * Tests a forms validation after all validation checks have responded
    * Polls validations responses every 50ms to allow uniqueness AJAX calls to complete
    */
-  testValidation: function($this) {
+  testValidation: function($this, $scope) {
```
tshedor: This `$scope` argument seems misnamed, unless it's used for another feature in the pipeline that I'm unaware of. For better maintainability, it should read as a boolean like it's used, i.e. `shouldReturnToTopOnSuccess`. In the function you're adding this to (`formValidate`), it will always be `true`; the value of the variable doesn't need to be passed on.

app/views/fae/application/_file_uploader.html.slim
```diff
@@ -28,7 +28,10 @@ ruby:
         .asset-actions.-files
           a.asset-title href=the_file.asset.url target="_blank" = the_file.asset.file.filename
 
-          = link_to '', fae.delete_file_path(the_file.id), class: 'asset-delete js-asset-delete', remote: true, method: :delete, data: { confirm: t('fae.delete_confirmation') }
+          - if the_file.id.present?
+            = link_to '', fae.delete_file_path(the_file.id), class: 'asset-delete js-asset-delete', remote: true, method: :delete, data: { confirm: t('fae.delete_confirmation') }
+          - else
+            a.js-file-clear.asset-delete.js-asset-delete
```
tshedor: `js-file-clear` and `js-asset-delete` will be misleading UI because the user will think they're deleting (the JS in `_ajax.js` doesn't know better, it's just DOM manipulation). I vote to **not** show a delete button if the file didn't save.

tshedor: Holla @jamesmk

app/assets/javascripts/fae/form/_ajax.js
tshedor: And - avoid [qualifying elements](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#qualifyingelement) like `a.`

app/assets/javascripts/fae/form/_ajax.js
```diff
@@ -78,6 +79,9 @@ Fae.form.ajax = {
       Fae.form.text.initMarkdown();
       Fae.form.checkbox.setCheckboxAsActive();
 
+      // validate nested form fields on submit
+      Fae.form.validator.formValidate($('.nested-form'));
```
tshedor: This jQuery element is referenced earlier in the closure - it should be cached and hoisted if it's being called twice.

tshedor:
Hi @jclemans hello it's me, the ghost of Frob-tend Past. You'll be visited by three spirits tonight, each encouraging you to refactor and correct our sins.
Good luck, we're all counting on you

app/assets/javascripts/fae/form/_validator.js
jclemans: It's not pretty or elegant. but I needed a convenient way to handle cases where $scope is the normal document body and cases where it is an AJAX loaded nested form. The latter was previously being completely ignored by these functions and thus providing no front end validation of an AJAX loaded nested form. If there are form errors in your nested form, then scrolling all the way to the top of the parent form is not an ideal user experience. I can take a stab at renaming and refactoring this though.

app/views/fae/application/_file_uploader.html.slim
jclemans: Given you're in a form error state - Without a delete button on the form the user is either forced to refresh the page to clear out the selected file or to correct form errors and submit the currently chosen file, and then delete it and choose a different file. Those don't seem like good user experiences if you accidentally select the wrong file or change your mind mid process.

app/assets/javascripts/fae/form/_ajax.js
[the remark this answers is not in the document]
jclemans: My intention was to get the nested form rendering working when there are errors with the least amount of changes to the code.

PR #422, opened 2018-08-28
Adding save_button param on _form_header.html.slim for the option to …

andregr15:
…hide the save_button on forms.

As the parameter to hide the new button already in the _index.html.erb file, I created a parameter to hide the save button of the form.

### DIFF
--- a/app/views/fae/shared/_form_header.html.slim
+++ b/app/views/fae/shared/_form_header.html.slim
@@ -1,5 +1,6 @@
 ruby:
   require_locals ['header'], local_assigns
+  save_button           = !local_assigns[:save_button].eql?(false)
   save_button_text    ||= 'Save'
   cancel_button_text  ||= 'Cancel'
   cloneable           ||= false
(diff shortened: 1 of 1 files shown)


jamesmk:
@andregr15 Thanks for the PR! I curious what your use case is to hide the save button on the form? I'd like to keep as many built in levers as low as possible to help keep the codebase maintainable. If it's an edge case, it would be easy to hide the button with CSS (you'll probably want to hide the cancel btn too).

andregr15:
Hello @jamesmk, i use to show a sended email list.

In my case i made a e-mail marketing system for study and in the e-mail has a confirmation pixel, when the user open the e-mail, a request is made to my site and that e-mail is marked as opened, so i made this change to the user access the e-mail as a read only resource and changed the name of the cancel button to go back to the list of e-mails.

jamesmk:
This seems like a misuse of the edit form. There are are other ways to submit the form without a submit btn which leaves this feature half baked. For this use case you may consider adding a custom show view to your admin controller. Or you can hack it by hiding the save btn with css and preventing form submit with JS.

thanks

Issue #409, opened 2018-05-26
Can't update a user, client validation issue


twig2let:
Brand new install v2.0.0.

Attempt to edit a user, http://0.0.0.0:3000/admin/users/1/edit

On save validation script throws exception: Uncaught TypeError: Cannot read property 'allow_blank' of undefined

The Password input data-validate tag contains JSON, so the Judge.ValidationQueue method fails.

data-validate="["{\"kind\":\"length\",\"options\":{\"minimum\":8,\"allow_blank\":true},\"messages\":{\"too_short\":\"is too short (minimum is 8 characters)\"}}"]"

It appears to be the only input whose data-validate that isn't an Object. 

Have I overlooked something? 

Rails 5.0.7
ruby 2.5.1p57
Fae 2.0.0

jamesmk:
Thanks for the report @twig2let. I am able to recreate the issue and will look into this week.

jamesmk:
Sorry for the delay. It appears the `data-validate` attribute is coming back from Judge (Fae's FE validation dependency: https://github.com/joecorcoran/judge) malformed on password inputs. Resolving bugs in dependencies can take a little more time as we'll have to wait for an official release of the Judge gem, then update the gemspec with the new version.

I'm working on a Fae CMS this week and will use the opportunity to look into more. If nothing else, I may submit a workaround until this is fixed in Judge.

thanks

mbajur:
Any dirty-hack to temporary fix this? :)

jclemans:
@mbajur - a quick hack I just used was to disable JS in my browser, make the edit to my account, then turn JS back on. You could also add a little JS snippet to remove that data-validate attr.

jamesmk:
Piggy backing off of what @jclemans was saying, here's some JS you can add to `fae.js` to disable JS validation on password fields:

```js
$(function() {
  $('input.password').data('validate', '');
});
```

The Rails validations will still be in place (> 8 chars and matching confirmation), they just won't happen via JS.

PR #426, opened 2018-09-11
Issue #409 - fix bugs in stripValidation()

aleksb:
Fixes issue #409 

In my case this issue manifested in being able to update users only in Firefox, but not Chrome or Edge.

Two bugs were fixed:

- Deleting an item from the array "skips over" the next item because the array gets shortened
- jQuery.attr leads to double-stringifying of objects which Judge chokes on (but apparently not in Firefox)


jamesmk:
Thanks for the PR! I'll verify the bugs fixed soon and get this thing merged.

PR #433, opened 2018-10-10
Fixes #432 - Create association on locale change

### LINKED ISSUE #432: Locale is being ignored for content blocks with multi lang support
What I have in `I18n.locale` is being ignored and the first/default language content is always fetched. 

Rails 5.1, fae-rails 2.0.0

```ruby
# config/initializers/fae.rb

Fae.setup do |config|
  config.languages = {
    en: 'English',
    cs: 'Česky'
  }
end
```

```ruby
# app/controllers/static_pages_controller.rb
class StaticPagesController < ApplicationController
  before_action :set_locale

  def home
    @item = HomePage.instance
  end

  private

  def set_locale
    I18n.locale = params[:locale] ||
                  request.env['HTTP_ACCEPT_LANGUAGE'].scan(/^[a-z]{2}/).first.presence ||
                  I18n.default_locale
  end
end
```

```ruby
# app/models/home_page.rb
class HomePage < Fae::StaticPage

  @slug = 'home'

  # required to set the has_one associations, Fae::StaticPage will build these associations dynamically
  def self.fae_fields
    {
      main_title: { type: Fae::TextField, languages: Fae.languages.keys },
      subtitle: { type: Fae::TextField, languages: Fae.languages.keys }
    }
  end
end
```

The locale is set up correctly, but I always get the incorrect content:

It must be some kind of noob error, could anyone point me to what I am doing wrong please?

xprazak2:
According to the docs, calling MyPage.instance.header
should get the header text for current locale.
However, the associations are set up in a singleton
for a default locale and never modified.

The language-specific methods like MyPage.instance.header_en
still work, but it would be nice to have fae handle the locale
for us.

This change checks for languages in fae_fields for page
and dynamically changes the association if the locale change
is detected.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ gem 'sass', require: 'sass'
 group :test, :development do
   gem 'rspec-rails', '~> 3.6.0'
   gem 'pry'
+  gem 'pry-byebug'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -92,6 +92,7 @@ GEM
       debug_inspector (>= 0.0.1)
     browser (2.5.3)
     builder (3.2.3)
+    byebug (10.0.2)
     capistrano (3.10.1)
       airbrussh (>= 1.0.0)
       i18n
@@ -226,6 +227,9 @@ GEM
     pry (0.11.3)
       coderay (~> 1.1.0)
       method_source (~> 0.9.0)
+    pry-byebug (3.6.0)
+      byebug (~> 10.0)
+      pry (~> 0.10)
     public_suffix (3.0.2)
     rack (2.0.4)
     rack-test (1.0.0)
@@ -368,6 +372,7 @@ DEPENDENCIES
   guard-rspec (~> 4.7.3)
   mysql2
   pry
+  pry-byebug
   rails (~> 5.2.0)
   rails-controller-testing
   rails-perftest
@@ -381,4 +386,4 @@ DEPENDENCIES
   yarjuf
 
 BUNDLED WITH
-   1.13.0
+   1.16.5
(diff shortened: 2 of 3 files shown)


xprazak2:
There is no way I can reasonably satisfy code climate - if I use the guard clause, then I get flagged for the line length.

johnnyshields:
I would suggest to remove the singleton altogether. Does it really add a substantial performance benefit?

xprazak2:
I removed the singleton as suggested.

xprazak2:
Fixing tests, can't get around the line length/guard clause rules combo.

jamesmk:
@xprazak2 I'm sorry if the docs weren't clear, but auto-translating attrs is provided by the `fae_translate` method (https://github.com/wearefine/fae/blob/65a4cb2304969e0ffcf58290b7b3cd0a415c7e69/app/models/concerns/fae/base_model_concern.rb#L66). This method doesn't currently support `Fae::StaticPage`.

The `@singleton_is_setup` is crucial for performance reasons, otherwise all the methods would have to be redefined every time you request an instance. So the meta programming can't change based on `I18n.locale`.

However, it would be possible to omit the base attr name here and then layer on a dynamic method similar to `fae_translate`. So `Fae::StaticPage` would define `header_en` and `header_cs` and layed on top would be something like:

```ruby
translated_attributes.each do |attribute|
  define_method attribute.to_s do
    self.send "#{attribute}_#{I18n.locale}"
  end
end
```

I'm going to close out this and open a ticket. I do like the refactoring  you made in the model, so if you want to include that or and attempt at the above, I'll make it a priority.

thanks!

xprazak2:
Adding `fae_translate :field_name` to static page as described in https://github.com/wearefine/fae/pull/472 works flawlessly, thanks!

PR #434, opened 2018-10-13
Fixes #240 - Add I18n support for admin interface

### LINKED ISSUE #240: Internationalization limitations
Hi!

I recently started using Fae in a new project and I'm missing some internationalization functionalities. In my project I needed to override the "set_locale" private method in Fae's ApplicationController to set the locale with the configured default locale in my rails app. Also there are many terms in the views that do not use I18n, like the links in application/_header partial.

xprazak2:
This adds the possibility of having admin iterface
translated. The language is chosen based
on the language of the user that is currently
logged in. If no user is logged in,
the default is English.

Note that this does not interfere with the actual
language of the site when the site ApplicationController
sets the locale from params as usual.

I added the translations for Czech.


xprazak2:
I added versions to the migrations in dummy app for backward compatibility to be able to run tests locally. Travis did not catch it as the database is created from schema and migrations are not executed.

I am not sure what to do with the remaining codeclimate issues - the unescaped attribute seems like a false positive to me since we render the `filter_select` and the `unless !` was already there, comment says it should be like that.

jamesmk:
Thank you @xprazak2 for this amazing PR! Sorry for the delay, but I will look this over as soon as I have availability.

johnnyshields:
Awesome! Will contribute a Japanese translation once this is merged.

jamesmk:
@xprazak2 Sorry for the delay. I'd like to include this awesome feature in the 2.1 release. I'll merge the PR into there and run another round of acceptance tests.

PR #442, opened 2018-12-19
77615 fixes for errors on activity log page and associated filtering

jclemans:
Activity log page is throwing errors for multiple issues addressed here.

* The /filter route was broken because it did not match up to the path in the xhr post, so we were just getting a 404 on every filter query result.
* Moved the change_item_link method to the change model to fix a TypeError
* date_scope was always nil because we were not comparing the strings with escaped space chars (i.e. `Last%20Week`)
* Removed the Arel.sql method used in the filtering query because it can't handle nils or empty arrays when there are unselected params. We can just do a straight up where(conditions) instead here.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,7 +15,7 @@ Fae::Engine.routes.draw do
   get 'settings' => 'users#settings', as: 'settings'
   get 'help' => 'pages#help', as: 'help'
   get 'activity' => 'pages#activity_log', as: 'activity_log'
-  post 'filter' => 'pages#activity_log_filter', as: 'activity_log_filter'
+  post 'activity/filter' => 'pages#activity_log_filter', as: 'activity_log_filter'
 
   get 'first_user' => 'setup#first_user'
   post 'first_user' => 'setup#create_first_user'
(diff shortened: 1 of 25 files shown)


config/routes.rb
```diff
@@ -15,7 +15,7 @@
   get 'settings' => 'users#settings', as: 'settings'
   get 'help' => 'pages#help', as: 'help'
   get 'activity' => 'pages#activity_log', as: 'activity_log'
-  post 'activity/filter' => 'pages#activity_log_filter', as: 'activity_log_filter'
```
jclemans: This route was not matching up with the POST request generated in `_filtering.js`, so all filters were resulting in a 404 error.

jclemans:
@jamesmk - I found some time to do address the issues and get this branch working right today. Seems to be working well now.

jclemans:
Just resolved conflicts from @jasonfine recent changes that were merged into master. Should be back in a good state on this branch now.

jclemans:
The diff on this seems to be messed up after rebasing with master. Going to close and open a fresh PR for this fix.

PR #444, opened 2018-12-28
Fix languages selector attribute in shared form headers documentation

Shigawire:
The current documentation errorneously describes the language switcher flag to be enabled with the option `language: true`. However, it would be correct to use `languages: true`.


jamesmk:
Thanks for the correction @Shigawire!

PR #469, opened 2019-04-25
locale preferences from browser

jamesmk:
Updates some changes from #434: `set_locale` based on browser language preferences. The language attribute on `Fae::User` is tied to the language content editor in Fae and can be independent of the preferred interface language.

Use case: English speaking content editor is responsible for all Japanese translations. They only want to see Japanese content fields but would prefer the admin interface to be English.

cc @xprazak2 @johnnyshields, I know you're both invested in this feature 🙂


jamesmk:
I'm merging this as the v2.1 branch is broken without this fix. I'm still open to feedback before 2.1 is officially released.

xprazak2:
Works for me, it makes sense to use the language from browser preferences :+1:

PR #475, opened 2019-05-08
GraphQL support

jamesmk:
- Fae generators will create GraphQL types if app uses GraphQL
- documents GraphQL setup


jamesmk:
Failing tests are from a previous merge on the v2.1 branch. I'm going to fix those directly on v2.1.

PR #480, opened 2019-06-07
Documentation Fix - Getting Content Blocks

### LINKED ISSUE #431: Documentation bug - Getting Content Blocks
Hello! 

I noticed in the "Getting Content Blocks" section of the "Pages and Content Blocks" documentation ([closest anchor link](https://www.faecms.com/documentation/topics-pages#adding-content-blocks)), alt text and caption for images is being accessed through the `asset` object.

screen shot 2018-10-02 at 10 56 05 am

I think this is incorrect, as I am unable to access those data points through the `asset` object (looks like this is the image uploader), but I _can_ access them through the field (`Fae::Image` object) directly, like so:

```ruby
@page.hero_image.alt
@page.hero_image.caption
```

I am using `fae-rails v2.0.0`.

Is the documentation incorrect? If so, I'd be more than happy to open up a PR to update.

PS: Looks like the "Getting Content Blocks" doesn't have an entry in the table of contents on that page either. I'd be glad to add that in as well while I'm in here.

harimohanraj89:
This PR fixes #431 by doing the following:

* Fix image `alt` and `caption` documentation
* Add missing entry to table of contents

### DIFF
--- a/docs/topics/pages.md
+++ b/docs/topics/pages.md
@@ -120,8 +120,8 @@ Then to get content from a `Fae::TextField` and `Fae::TextArea`:
 # for `Fae::Image` or `Fae::File`
 @about_us_page.hero_image.asset.url
 # for `Fae::Image` only
-@about_us_page.hero_image.asset.alt
-@about_us_page.hero_image.asset.caption
+@about_us_page.hero_image.alt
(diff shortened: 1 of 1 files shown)


jamesmk:
Thanks for the PR @harimohanraj89

PR #488, opened 2019-07-08
80330 security fixes

jamesmk:
**Prevent vertical privilege escalation**
Users with the role of admins could potentially set their role as super admin by abusing the `toggle` method provided for boolean attributes. Stricter definitions were added to `toggle` to make sure this doesn't happen. 

**More toggle abuse**
`toggle` could also be used to on non-boolean attributes, which could lead to other potential security issues. Method is now available for booleans only.

`toggle` also exposed the existence of classes but returning a 500 when a class did not exist. Missing classes now return a 401 inline with unauthorized classes.

**Brute-force protection**
Currently Fae has no brute-force protection. This also enables Devise's lockable feature which will lock accounts after 5 failed attempts to log in. Accounts unlock after an hour or via emailed instructions.

### DIFF
--- a/app/controllers/fae/users_controller.rb
+++ b/app/controllers/fae/users_controller.rb
@@ -23,6 +23,8 @@ module Fae
     end
 
     def create
+      authorize_role
+
       @user = Fae::User.new(user_params)
 
       if @user.save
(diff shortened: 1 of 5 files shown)


jasonfine:
sweet!

PR #490, opened 2019-07-31
fix nested form redraws from showing raw html instead of form

jasonfine:
this fixes the issue where raw html is getting rendered instead of the form when submitting a nested form with image uploaders that end up failing validation


jasonfine:
will fix that failing test when i get time

PR #494, opened 2019-09-15
fix typos on empty home page

pjanek:


jamesmk:
Thanks for the PR @pjanek!

Issue #496, opened 2019-09-19
Disable "Add" button option on nested_table


jclemans:
A request keeps coming up to add a nested table to a form page and to disable the ability to add a new object here. Sometimes the nested objects are imported or associated in some other way. Adding a custom table view instead of using the fae nested table works as a workaround, but it would be super handy to just be able to `render 'fae/shared/nested_table', add_button: false` and move on to other things.

jclemans:
Fix merged in!

PR #497, opened 2019-09-20
Optional add button for nested tables

jclemans:
Related to https://github.com/wearefine/fae/issues/496

Found that this was a pretty simple change to add an optional argument that can be used to disable the add button and the "add some" link on a nested table. This should be a lot faster and easier to use than creating a custom table just to override the default behavior.


jasonfine:
🎉 shipit!

PR #498, opened 2019-09-24
77615 activity log fixes

jclemans:
Creating a fresh PR for this issue after updating the branch and rebasing. This branch fixes Fae activity log page errors that were preventing filtering from working.

* Normalizes the date_scope because it was always getting set to nil. We were not comparing the strings with escaped space chars (i.e. Last%20Week never matches "Last Week")
* Removes the hidden class from the reset button so filters can be reset after a search


jasonfine:
👏

jclemans:
@jamesmk - Any chance we can get this change merged in? This is still an open issue for production sites running the current Fae version.

PR #541, opened 2020-09-15
Adds highlight color admin control in root settings

jclemans:
Fae admins rejoice! For this change allows the highlight color to be changed in the root settings instead of requiring a dev to manually update the fae.scss file with a new hex color.

Note, after incorporating this change you will need to run `rake fae:install:migrations` and then run `rake db:migrate` to get the new column into your app's db.

Related to: FINE Redmine issue 77928

### DIFF
--- a/app/assets/stylesheets/fae/globals/_tags.scss
+++ b/app/assets/stylesheets/fae/globals/_tags.scss
@@ -33,7 +33,7 @@ body {
 
 a {
   @include transition(color .2s);
-  color: $c-custom-highlight;
+  color: var(--highlight-color);
   text-decoration: none;
 
   &:hover {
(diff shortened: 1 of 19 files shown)


jamesmk:
We will be including another feature in Fae 2.2 that requires `rake fae:install:migration` so we should bundle these features.

Nice work on the dynamic injection of CSS styles and on the fly `precompile`. i am hesitant to use that approach because I fear it will break down in a variety of hosting environments.

It looks like dynamic CSS vars defined at the markup level are now well supported and I think that would be a more stable approach. https://ishadeed.com/article/css-variables-inline-styles/

We should be able to inject this new managed hex in the layout and use that var through Fae's SCSS:

```slim
body class=body_class style="--custom_highlight: #88bbee"
```

```css
color: var(--custom_highlight, [fallback]);
```

jamesmk (changes requested):
https://github.com/wearefine/fae/pull/541#issuecomment-1023535325

jamesmk:
Closing in favor of #572

PR #543, opened 2020-10-09
Generator fixes for GraphQL Types and some template spacing fixes

jclemans:
A few fixes for things I noticed weren't working quite right when generating new objects or pages:

* New page types now inherit from `BaseFaePageType`
* Types ignore the `:static_page` attr so it doesn't get added to the type file and break things.
* Templates indentation containing tabs is converted to spaces for consistency and better readability

### DIFF
--- a/lib/generators/fae/templates/graphql/graphql_type.rb
+++ b/lib/generators/fae/templates/graphql/graphql_type.rb
@@ -5,7 +5,9 @@ class Types::<%= class_name %>Type < Types::BaseObject
   field :id, ID, null: false
 <% if @graphql_attributes.present? -%>
 <% @graphql_attributes.each do |graphql_object| -%>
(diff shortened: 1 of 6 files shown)


lib/generators/fae/templates/graphql/graphql_type.rb
```diff
@@ -5,8 +5,10 @@ class Types::<%= class_name %>Type < Types::BaseObject
   field :id, ID, null: false
 <% if @graphql_attributes.present? -%>
 <% @graphql_attributes.each do |graphql_object| -%>
+<% if graphql_object[:attr] != :static_page -%>
```
jamesmk, line 8: The above suggestion would also nullify this change, because without `BaseFaePageType` we'd want to keep the the auto-generation of static page GQL.

jamesmk (changes requested):

PR #550, opened 2021-02-16
Update to newer rails versions

cassianoblonski:


jamesmk:
Thanks for you contribution @cassianoblonski! We'll include this in our 2.2 release.

PR #573, opened 2022-08-09
94637 error message update

jaybones90:


app/views/fae/shared/_errors.slim
```diff
@@ -1,4 +1,4 @@
-- unless flash.present?
-  .flash-message.alert.hide = t('fae.save_error')
+/ - unless flash.present?
```
jamesmk, line 1: Remove commented out code

lib/judge/controller.rb
```diff
@@ -0,0 +1,58 @@
+require 'uri'
```
jamesmk, line 1: Let's discuss a PR into Judge if needed

jamesmk (changes requested):

PR #591, opened 2022-11-16
Add base MFA (through otp QRCodes)

jsonjordan:


app/controllers/fae/application_controller.rb
```diff
@@ -105,6 +108,12 @@ def first_user_redirect
       redirect_to fae.first_user_path if Fae::User.live_super_admins.blank?
     end
 
+    def setup_mfa_redirect
+      if @option.site_mfa_enabled && current_user.present? && current_user.user_mfa_enabled && !current_user.otp_required_for_login && request.path != new_two_factor_settings_path && request.path != two_factor_settings_path
```
jsonjordan, line 112: I hate this line, but its accurate to when we should redirect.  It forces the user to set up their MFA before they can go to any other page.  Would be super open to another way of getting this done.

app/models/fae/user.rb
```diff
@@ -1,13 +1,17 @@
...
+
     # Include default devise modules. Others available are:
     # :registerable, :confirmable, :timeoutable and :omniauthable
-    devise :database_authenticatable,
-           :recoverable, :rememberable, :trackable, :lockable
+    devise :recoverable, :rememberable, :trackable, :lockable
+
+    devise :two_factor_authenticatable, :two_factor_backupable,
+      otp_backup_code_length: 10, otp_number_of_backup_codes: 10,
+      :otp_secret_encryption_key => ENV['OTP_SECRET_KEY']
```
jsonjordan, line 14: Reminder:  Need to add this ENV var to setup docs

app/views/fae/two_factor_settings/edit.html.erb
```diff
@@ -0,0 +1,33 @@
+<div class="container">
```
jsonjordan, line 1: Needs to be changed into slim.  Could use some help on that part.

app/views/fae/two_factor_settings/new.html.erb
```diff
@@ -0,0 +1,42 @@
+<div class="container">
```
jsonjordan, line 1: Needs to be changed into slim.  Could use some help on that part.

app/controllers/fae/two_factor_settings_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      if current_user.two_factor_backup_codes_generated?
+        flash[:alert] = 'You have already seen your backup codes.'
+        return redirect_to users_path
+      end
+
+      @backup_codes = current_user.generate_otp_backup_codes!
+      current_user.save!
+    end
+
+    # def destroy
```
jsonjordan, line 46: This function is not used, as we destroy on a toggle on the users table.  I felt keeping it around for reference was worth it.

PR #596, opened 2023-05-29
Update page_generator.rb

losingle:
Ruby 3.0.0 removing File.exists?

### DIFF
--- a/lib/generators/fae/page_generator.rb
+++ b/lib/generators/fae/page_generator.rb
@@ -26,7 +26,7 @@ module Fae
 
     def generate_static_page_controller
       file = "app/controllers/#{options.namespace}/content_blocks_controller.rb"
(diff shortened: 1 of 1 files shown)


MrSkinny:
FYI, without this being merged, I was following your [tutorial](https://www.faecms.com/documentation) and the reader is unable to progress past the instructed command of `rails g fae:page AboutUs hero_image:image headline:string body:text` as it returned an undefined method exception:

```
~/.rbenv/versions/3.2.3/lib/ruby/gems/3.2.0/gems/fae-rails-3.0.0/lib/generators/fae/page_generator.rb:29:in `generate_static_page_controller': undefined method `exists?' for File:Class (NoMethodError)
```

jamesmk:
Thank you for the PR, we'll include this in our 3.1 release!

PR #599, opened 2023-09-20
Language transcription feature

jsonjordan:
Still working on an unrelated bug that affects this feature, but wanted thoughts and opinions!


spec/spec_helper.rb
```diff
@@ -5,6 +5,7 @@
 require 'capybara/rspec'
 require 'capybara-screenshot/rspec'
 require 'yarjuf'
+require 'factory_girl'
```
jasonfine, line 8: See previous note about factory_girl

app/helpers/fae/form_helper.rb
```diff
@@ -139,6 +141,13 @@ def label_and_hint(attribute, options)
       options[:hint] = hint.html_safe if hint.present?
     end
 
+    def translate_button(f, attribute, options)
+      if Fae::Option.instance.translate_language && !attribute.to_s.include?('_en') && Fae.languages.keys.any? { |lang| attribute.to_s.include?(lang.to_s) }
+        translate = content_tag(:span, 'Translate', class: 'button js-translate-button', style: 'position: absolute; margin-left: 10px;')
```
jasonfine, line 146: Plz make this styling a CSS class instead of inlining it here. There's a couple more instances of this happening that can be addressed as well

app/controllers/fae/utilities_controller.rb
```diff
@@ -61,6 +70,32 @@ def can_toggle(klass, attribute)
       true
     end
 
+    def translate_request(language, en_text)
+      subscription_key = ENV['TRANSLATOR_TEXT_SUBSCRIPTION_KEY']
+      endpoint = ENV['TRANSLATOR_TEXT_ENDPOINT']
+      path = '/translate?api-version=3.0'
```
jasonfine, line 76: Since it seems like this is built on the assumption that a specific service provider will be used, i don't think we need to hide the endpoint in an ENV var, one less thing to chase around as an ENV var.

Gemfile
```diff
@@ -30,6 +30,7 @@ end
 
 group :test do
   gem 'webrick'
+  gem 'factory_girl', '~> 4.9'
```
jasonfine, line 33: Curious about this, `factory_girl` was renamed a while back to `factory_bot` and shouldn't the following line for `factory_bot_rails` satisfy the app's needs?

jasonfine (commented):
Overall looking great, a few things noted inline.

Is this working for images in Fae pages?

We'll want to think about some feature tests for this. Theres already a bunch of form feature tests in the app so we can probably add these there.

We don't need to test that we're getting a response off of the API, but we'll need to capture an API response to use as a mock for tests, so we can fully test our implementation of the returned data. HMU if you want to chat about a strategy here.

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -56,6 +57,38 @@ Fae.form.text = {
     $('.js-markdown-editor:not(.mde-enabled)').each(function () {
       var $this = $(this);
 
+      // set up translate button for markdown fields and add helper text
+      var translate_button = $this.siblings( '.js-translate-button' )
+      if (translate_button.length) {
```
jaybones90, line 62: the `length` property on the jquery object will still be present even if it doesn't find the element so this should check to ensure `length > 0`

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -56,6 +57,38 @@ Fae.form.text = {
...
+        var label = $this.siblings('.has_no_helper_text')
+        if (label.length) {
```
jaybones90, line 64: same here

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -56,6 +57,38 @@ Fae.form.text = {
...
+          jQuery('<span>', {
+            class: 'helper_text_text',
+            text: 'If machine translation is required, please click “Translate” button before applying any text formatting in markdown.'
+          }).appendTo(helper_class);
+
+          helper_class.appendTo(label)
+        }
+
+        var label_height = $this.siblings( 'label' ).height()
+        $this.siblings( '.js-translate-button' ).closest(".input").css({
```
jaybones90, line 78: let's keep the CSS in stylesheets instead of within the JS. you can additional classes here via JS and then target those classes in a css file. `/app/assets/stylesheets/fae/modules/forms/_base.scss` would probably be a good place

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -56,6 +57,38 @@ Fae.form.text = {
     $('.js-markdown-editor:not(.mde-enabled)').each(function () {
       var $this = $(this);
 
+      // set up translate button for markdown fields and add helper text
+      var translate_button = $this.siblings( '.js-translate-button' )
```
jaybones90, line 61: A standard convention when working with jquery is to prepend a `$` sign to any variables that are jquery objects. This helps to differentiate between jquery vs. non-jquery variables at a glance. Let's also use camelCase when working with JS so `translate_button` should become `$translateButton`

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -128,5 +164,71 @@ Fae.form.text = {
...
+   * Find all translate fields and initialize them with button
+   */
+    initTranslation: function () {
+      // var $translate_button = $('.js-translate-button');
+      $('.js-translate-button').click(function(e) {
+        var $this = $(this);
+        var translate_field;
+
+        // grab the field the button belongs to
+        if (this.closest(".text")) {
```
jaybones90, line 179: this check should be updated to `this.closest(".text").length > 0` to ensure it's actually found before manipulating

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -128,5 +164,71 @@ Fae.form.text = {
...
+          translate_field = this.closest(".text");
+        } else  {
+          translate_field = this.closest(".string");
+        }
+
+        // grab language and model name from field, container logic is for image alt text
+        var translate_language = translate_field.attributes['data-language'].value;
+        var translate_model
```
jaybones90, line 187: missing semicolon

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -128,5 +164,71 @@ Fae.form.text = {
...
+        if (translate_language.length == 4) {
+          translate_language = `${translate_language.slice(0,2)}-${translate_language.slice(2)}`
+        }
+
+        $.ajax({
+          url: Fae.path + '/translate_text',
+          type: "post",
+          data: { translation_text: { language: translate_language, en_text: english_text } },
+          success: function(data) {
+            // set translation text into tranlate model
```
jaybones90, line 219: typo in comment

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -128,5 +164,71 @@ Fae.form.text = {
...
+            if ($this.siblings('.CodeMirror').length) {
```
jaybones90, line 220: check if `length > 0`

app/assets/javascripts/fae/form/inputs/_text.js
```diff
@@ -128,5 +164,71 @@ Fae.form.text = {
...
+        // get translate_language in correct format for request
+        if (translate_language.length == 4) {
+          translate_language = `${translate_language.slice(0,2)}-${translate_language.slice(2)}`
+        }
+
+        $.ajax({
+          url: Fae.path + '/translate_text',
+          type: "post",
+          data: { translation_text: { language: translate_language, en_text: english_text } },
+          success: function(data) {
```
jaybones90, line 218: we should probably have a check to ensure the `data` being returned from the response exists and is in the format you expect before continuing on any further in the function. since your accessing it like `data[0]`, something like `if (data && data.length > 0)` or whatever seems appropriate

spec/spec_helper.rb
jsonjordan: updated!

app/helpers/fae/form_helper.rb
jsonjordan: updated!

app/controllers/fae/utilities_controller.rb
jsonjordan: updated!

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!

jsonjordan: Though a few I tried ended up breaking stuff.  Next PR, if you still think some vars should be renamed with the $, we can get on a call!

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!  (I am a typo machine)

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!

app/assets/javascripts/fae/form/inputs/_text.js
jsonjordan: updated!

jsonjordan:
If you can also look for typos on the docs.  I found a few and fixed them, but am not confident I got everything.

jsonjordan:
@jamesmk I also am unsure about the CircleCI Pipeline issue, I cant see a config.yml file in my branch or master.  🤷‍♂️

PR #605, opened 2024-03-19
Slack notify integration

jsonjordan:


Gemfile.lock
```diff
@@ -36,6 +36,7 @@ PATH
       sass (>= 3.4.0)
       sass-rails (>= 5.0.7)
       simple_form (<= 5.1)
+      slack-notifier
```
jasonfine, line 39: Add this to the gemspec file

app/models/concerns/fae/base_model_concern.rb
```diff
@@ -2,12 +2,38 @@ module Fae
...
       include Fae::Trackable if Fae.track_changes
       include Fae::Sortable
+      after_create :notify_initiation
+      before_save :notify_changes
+    end
+
+    def notify_changes
+      return unless notifiable_attributes.present?
+      notifiable_attributes.each do |field_name_symbol|
+        if self.send("#{field_name_symbol}_changed?") && self.send(field_name_symbol).present?
```
jasonfine, line 19: Confirming we don't want to notify if a field value was deleted?

jsonjordan: Not at this time, but def something that could come up in future

jasonfine:
Docs added @jamesmk

PR #607, opened 2024-04-19
Open nested forms in context

jasonfine:
Nested forms now open under the row being edited.

Adding new opens the row under the table header.

'Double nestings' are supported, as seen with Aromas and Sub Aromas here https://fae-dummy.afinesite.com/admin/releases/1/edit.

All tests are passing except for the few that seem to flicker across all feature branches.


jasonfine:
@jamesmk up on dev, the Aromas/SubAromas nesting here https://fae-dummy.afinesite.com/admin/releases/3/edit illustrates the changes, however do note that this is a global behavior change for all nested tables and not opt-in, so that's clear.

jasonfine:
@jamesmk Closing this for now, seeing some weirdness mixing this with the flex components, which i feel is a more important feature to get into the core so we can start iterating on UX improvements with respect to that feature.

PR #612, opened 2024-06-06
Flex components

jasonfine:
This follows the typical pattern of the core of the feature being in the Fae core and the installer adds some files to the mounting app. Originally the installer was adding the core flex component route and views into the mounting app. I removed this behavior and tucked all of that into the core as well so future UI improvements won't require updating generated files in the mounting apps.

The Dummy app is now GraphQL enabled and the settingslogic gem was added so it can be used similarly to how all of our builds are set up these days.

The documentation explains the rest https://github.com/wearefine/fae/blob/040c1f1589391198ac0652c5176b94d261a366b2/docs/features/flex_components.md

### DIFF
--- a/app/helpers/fae/nested_form_helper.rb
+++ b/app/helpers/fae/nested_form_helper.rb
@@ -29,7 +29,7 @@ module Fae
           end
         end
       elsif item.class.columns_hash[attribute.to_s].present? && item.class.columns_hash[attribute.to_s].type == :boolean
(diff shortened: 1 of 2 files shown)


jasonfine:
@jamesmk this can be seen here https://fae-dummy.afinesite.com/admin/content_blocks/components

Keep in mind that my goal here was getting the core data modeling and new generator into the core so we have a base to iterate on UX improvements. So on the view layer this will look/function identically to what we're used to now in our apps that use components.

